PR #11, opened 2019-04-01
Add go.mod

seblw:
Hello.
Adding such will allow folks to build project outside of the GOPATH.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/ffuf/ffuf
+
+go 1.11


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/ffuf/ffuf
+
+go 1.12
```
joohoi: Can we make this `go 1.11` or is it possible to leave it out altogether?

seblw: Removing wasn't possible as it auto generates, however downgrading it to go 1.11 works fine.

PR #12, opened 2019-04-01
Add output to CSV file

seblw:
Hi again,
Another PR this evening :)
Example output content:
```
input,status_code,content_length,content_words
ccc,301,0,1
aaa,301,0,1
```

I'm still not sure about adding timestamp or/and source file to the output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ To define the test case for ffuf, use the keyword `FUZZ` anywhere in the URL (`-
   -o string
     	Write output to file
   -of string
-    	Output file format. Available formats: json (default "json")
+    	Output file format. Available formats: json, csv, ecsv (default "json")
   -p delay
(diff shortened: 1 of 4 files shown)


seblw:
Yep, good point. Actually I think both approaches could be useful - maybe introducing two output formats like `ecsv` with base64 encoded payload and raw `csv` for some simpler use-cases like when you just bruteforce directories etc. That way the tool can support simple and advanced use cases (assuming that advanced user is able to decode the result hehe :) )

joohoi:
That sounds like a good plan! I'm all for it

seblw:
Good to hear! Will put an update around the evening.

seblw:
Updated.

joohoi:
LGTM! Thanks!

PR #18, opened 2019-04-06
clickable output

eur0pa:


eur0pa:
Or when `-e` is specified I guess? You do have a valid point, although the default use-case is forced-browsing

eur0pa:
Should be fine now

pkg/output/stdout.go
```diff
@@ -159,6 +163,11 @@ func (s *Stdoutput) resultNormal(resp ffuf.Response) {
 	fmt.Println(res_str)
 }
 
+func (s *Stdoutput) resultAlt(resp ffuf.Response) {
+	res_str := fmt.Sprintf("%s %-3s %-9d %-5d %s", TERMINAL_CLEAR_LINE, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords, resp.Url)
```
joohoi: I'd like to keep descriptive output in the default operation, so newcomers would be able to easily figure out which value is which. Alternatively a "table header" line for output would suffice with this kind of default output. I mean something like:

```
Status |   Size   | Words | Input
-----------------------------------
   201     1024        64   https://ffuf.io.fi/matched.result
   200     1315       124   https://ffuf.io.fi/another.result
```
and obviously when `FUZZ` keyword wasn't in the URL;

```
Status |   Size   | Words | Input
-----------------------------------
   201     1024        64   matched.result
   200     1315       124   another.result
```


Please let me know what you think. This doesn't have to happen in this PR, and I can implement that, but I'd like to hear your opinion nevertheless.

pkg/ffuf/response.go
```diff
@@ -12,6 +12,7 @@ type Response struct {
 	ContentLength int64
 	ContentWords  int64
 	Cancelled     bool
+	Url           string
```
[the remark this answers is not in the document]
eur0pa: But the `resp.Request.Url` contains the *target* URL, not the requested URL:

`req.Url = conf.Url`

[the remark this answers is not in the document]
joohoi: Yeah, you are right. That's only for the initialization though, it later gets replaced in the runner's `Prepare()` function with the parsed (replaced) request url: https://github.com/ffuf/ffuf/blob/master/pkg/runner/simple.go#L57

pkg/ffuf/config.go
```diff
@@ -54,6 +55,7 @@ func NewConfig(ctx context.Context) Config {
 	conf.StopOnErrors = false
 	conf.StopOnAll = false
 	conf.FollowRedirects = false
+	conf.AltOutput = false
```
joohoi: `AltOutput` doesn't exist in the config struct anymore, so this fails to compile.
```suggestion
```

pkg/output/stdout.go
```diff
@@ -146,7 +147,11 @@ func (s *Stdoutput) printResult(resp ffuf.Response) {
 	if s.config.Quiet {
 		s.resultQuiet(resp)
 	} else {
-		s.resultNormal(resp)
+		if s.config.AltOutput {
```
joohoi: AltOutput doesn't exist in the config struct anymore, so this fails to compile.
```suggestion
		if s.config.UrlOutput {
```

pkg/runner/simple.go
```diff
@@ -103,6 +103,7 @@ func (r *SimpleRunner) Execute(req *ffuf.Request) (ffuf.Response, error) {
 
 	wordsSize := len(strings.Split(string(resp.Data), " "))
 	resp.ContentWords = int64(wordsSize)
+	resp.Url = req.Url
```
joohoi: This isn't needed anymore
```suggestion
```

main.go
```diff
@@ -206,6 +206,7 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
 
 	//Search for keyword from URL and POST data too
 	if strings.Index(conf.Url, "FUZZ") != -1 {
+		s.config.UrlOutput = true
```
joohoi: This should probably be `conf.UrlOutput`
```suggestion
		conf.UrlOutput = true
```

pkg/output/stdout.go
```diff
@@ -155,7 +160,12 @@ func (s *Stdoutput) resultQuiet(resp ffuf.Response) {
 }
 
 func (s *Stdoutput) resultNormal(resp ffuf.Response) {
-	res_str := fmt.Sprintf("%s%-23s [Status: %s, Size: %d, Words: %d]", TERMINAL_CLEAR_LINE, resp.Request.Input, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords)
+	res_str := fmt.Sprintf("%s %3d   %-8d   %-5d   %s", TERMINAL_CLEAR_LINE, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords, resp.Request.Input)
```
joohoi: `colorizeStatus` returns `string` because it adds colors using terminal escape sequences.
```suggestion
	res_str := fmt.Sprintf("%s %3s   %-8d   %-5d   %s", TERMINAL_CLEAR_LINE, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords, resp.Request.Input)
```

pkg/output/stdout.go
```diff
@@ -155,7 +160,12 @@ func (s *Stdoutput) resultQuiet(resp ffuf.Response) {
...
+	fmt.Println(res_str)
+}
+
+func (s *Stdoutput) resultUrl(resp ffuf.Response) {
+	res_str := fmt.Sprintf("%s %3d   %-8d   %-5d   %s", TERMINAL_CLEAR_LINE, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords, resp.Request.Url)
```
joohoi: `colorizeStatus` returns `string` because it adds colors using terminal escape sequences.
```suggestion
	res_str := fmt.Sprintf("%s %3s   %-8d   %-5d   %s", TERMINAL_CLEAR_LINE, s.colorizeStatus(resp.StatusCode), resp.ContentLength, resp.ContentWords, resp.Request.Url)
```

Issue #19, opened 2019-04-06
[request] add arbitrary, multiple extensions to requests


eur0pa:
Flag: `-e`, Default: (none)
Commandline: `-e .aspx,.php,nodot`
Generator logic (FUZZ = current entry from wordlist): `/FUZZ /FUZZ.aspx /FUZZ.php /FUZZnodot`

joohoi:
Do you think it'd be feasible to list the extensions in a wordlist? This would allow more abstract way of doing this, that would at the same time enable different fuzzing modes with multiple FUZZ keywords. Something that Burp calls Pitchfork and Cluster bomb modes in the Intruder module.

eur0pa:
You normally don't need that many, it's usually 10 or less; commandline is practical enough especially when scripting the tool in automation chains, which would save a step of dumping extensions to file before invoking `ffuf`

C0RB3N:
Hey guys, I actually added this feature for a friend 🙂 It essentially replaces %EXT% like dirsearch does, so I'm sure it wouldn't be hard to just force extensions as well.

joohoi:
> Hey guys, I actually added this feature for a friend slightly_smiling_face It essentially replaces %EXT% like dirsearch does, so I'm sure it wouldn't be hard to just force extensions as well.

👋 I wonder if you would be up to submitting a PR?

C0RB3N:
Yep! I'll touch up the code a bit, it might be later this week as I'm going to the March Madness SemiFinals soon

joohoi:
Sounds good! Looking forward to it. I'd probably prefer some other keyword than %EXT%, though, FFUFEXT for example, even though I don't like that too much either ;) 

I'm up for discussion about keyword refactoring altogether.

C0RB3N:
I meant it replaces %EXT% from any given line in a wordlist :) (Dirsearch's wordlist contains %EXT% where it want's an extension bruteforced, was merely given the opportunity for users like [naffy](https://twitter.com/nnwakelam) use their desired wordlists. It also just force extensions if the line in the wordlist doesn't contain that keyword.

joohoi:
Yup, I was just meaning that as the main keyword is currently `FUZZ` to be replaced by wordlist entry, I would like some other keyword than `%EXT%` to be replaced by the extension provided in a list in command line, but am up for discussion about the keywords altogether.

All in all, I'd like the keyword to be easy to memorize, and something that wouldn't collide with actual requests or shell internals.

[...]

PR #23, opened 2019-04-06
Add -e flag for Extensions.

C0RB3N:
This adds an `-e` flag for extensions. This currently replaces %EXT% in a given wordlist, but we should soon add the option to append an extension to each line in the wordlist.

Example usage:

```bash
root@dev:~# ffuf -u http://example.com/FUZZ -e asp,html,jpg -w wordlist.txt

        /'___\  /'___\           /'___\
       /\ \__/ /\ \__/  __  __  /\ \__/
       \ \ ,__\\ \ ,__\/\ \/\ \ \ \ ,__\
        \ \ \_/ \ \ \_/\ \ \_\ \ \ \ \_/
         \ \_\   \ \_\  \ \____/  \ \_\
          \/_/    \/_/   \/___/    \/_/

       v0.9git
________________________________________________

 :: Method       : GET
 :: URL          : http://example.com/FUZZ
 :: Matcher      : Response status: 200,204,301,302,307,401,403
________________________________________________

:: Progress: [3/3] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
```

PR #25, opened 2019-04-10
-e flag to provide a list of extensions and -D for DirSearch wordlist format compatibility

joohoi:
This PR builds on work of @C0RB3N in #23 

Two flags are added in this PR:
`-e` that allows you to provide list of extensions to append to each of the wordlist entries.
For example, a wordlist with single entry : `example`, and `-e .first,.second,.third` would result into three different test cases: `example.first, example.second and example.third`.

`-D` flag to provide compatibility with DirSearch wordlists. Only the wordlist entries that contain `%EXT%` will get appended with extensions provided by `-e`

Fixes: #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ ffuf -w /path/to/postdata.txt -X POST -d "username=admin\&password=FUZZ" https:/
 
 To define the test case for ffuf, use the keyword `FUZZ` anywhere in the URL (`-u`), headers (`-H`), or POST data (`-d`).
 ```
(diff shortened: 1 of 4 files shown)


C0RB3N:
It might be easier for users to not have to specify the "." in front of the extension

joohoi:
> It might be easier for users to not have to specify the "." in front of the extension

You are right, but I opted in for greater versatility, this was also requested in the original issue that was created.

C0RB3N:
Yeah that's fair, I missed that in the original issue. Looks good.

joohoi:
Thanks for your work on the PR @C0RB3N !

Issue #4, opened 2019-03-20
Retry Connection Reset


fang0654:
First off, I love this tool. It is awesome. My only feature request: instead of printing `"Error in runner: Get <url> ...read: connection reset by peer"` and `"Client.Timeout exceeded"` errors, retry the attack and mark an 'error' counter with the status line. This will keep from dropping tests due to an overloaded server, and keep the output somewhat cleaner.

joohoi:
Thanks for your kind words, and for suggesting this feature. It's a really good idea, and I'll implement it soon(ish) when I get the time.

SolomonSklash:
I wanted to second this. I noticed that when it happens, the request per second drops very low, and delays the scan quite a bit. I use ffuf in a chain of tools via [Chomp Scan](https://github.com/SolomonSklash/chomp-scan), and when I hit this error, the entire script essentially hangs waiting for the timeouts and slow request rate.

Is there a way to do something similar to the `-sf` flag (great addition by the way!) and exit the scan if a lot of errors happen?

joohoi:
There are multiple things at play here, and I'm wondering about the approach to take. Currently, I'm thinking of adding a new flag `-se` (stop on errors) and `-sa` (stop on all, which would apply `-se` and `-sf` and probably others to come in the future).

For the retry, I'm wondering about safe defaults. I will be adding a new flag to do what @fang0654 suggested, but I'm wondering if there should be some logic alongside of it - like for example increasing the delay to avoid spurious connection errors, and whether to make these defaults or not.

I'm curious to hear about your opinions on this.

SolomonSklash:
I really like the `-se` and `-sa` flags. For my use case it's perfect.

For the use case where you don't want to stop if errors are encountered due to connection errors, increasing the delay sounds like a good idea. Making the delay optional and configurable could be good too, if it's not too much trouble to implement.

That's my two cents. Thanks for your work on this tool!

fang0654:
I'd say with the retries, only retry once. The main reason is that the error may be intentional on the client side. For example, I've seen a few WAFs that just drop the connection if your request matches some blacklist. I even had one recently that didn't bother to at least send a syn packet, so anything it considered malicious would just timeout. 

It seems like a tough trade-off of simplicity, efficiency and safety, although I guess having various flags for the different behaviours kind of takes care of that.

SolomonSklash:
Thanks for implementing this! Some feedback based on my (limited) initial testing:
[...]

PR #31, opened 2019-04-25
Add -timeout flag for customizable HTTP Request timeouts

lc:
Hi,

Like the title implies, this PR adds the command line flag `-timeout` which allows users to add a custom HTTP request timeout. I've had several use-cases for it recently so I figured I would add this.

By default, it is set to `10` seconds.


joohoi:
Thanks for the PR! It's definitely a good feature to have, we discussed it recently in #4 with @SolomonSklash as well.

lc:
Added!

main.go
```diff
@@ -78,6 +78,7 @@ func main() {
 	flag.BoolVar(&conf.FollowRedirects, "r", false, "Follow redirects")
 	flag.BoolVar(&conf.AutoCalibration, "ac", false, "Automatically calibrate filtering options")
 	flag.IntVar(&conf.Threads, "t", 40, "Number of concurrent threads.")
+	flag.IntVar(&conf.Timeout, "timeout", 10, "HTTP request timeout. Default: 10")
```
joohoi: The default is printed automatically in the CLI help so the `Default: 10` part can be removed.

joohoi: Oh, and we could mention that the unit is second, so something along the lines of
```
HTTP request timeout in seconds.
```

Though this is probably obvious for 99% of the users, it doesn't hurt to be more explicit.

lc: Done

lc:
Should be all good to go now

PR #36, opened 2019-06-04
Wordlist standard input mode

delic:
Added the possibility to read wordlist from standard input with special filename - like so:
seq 1 1000 | ffuf -w - -u ...

Also added the compiled binary to .gitignore, sorry.


README.md
```diff
@@ -143,8 +143,8 @@ The only dependency of ffuf is Go 1.11. No dependencies outside of Go standard l
       - New CLI flag: -ac to autocalibrate response size and word filters based on few preset URLs.
       - New CLI flag: -timeout to specify custom timeouts for all HTTP requests.
 
-
    - Changed
+      - Wordlist can be read also from standard input
```
joohoi: Grammar fix:

```suggestion
      - Wordlist can also be read from standard input
```

PR #47, opened 2019-06-27
Added support for using ranges in size, word count or status code matching/filtering

### LINKED ISSUE #45: Limit results with with http response size range
Is there a simple way to filter out responses with N <= size <= M ?

delic:
Resolves #45


pkg/filter/valuerange.go
```diff
@@ -0,0 +1,38 @@
+package filter
```
joohoi: I would love to see this moved to the main library: `pkg/ffuf`, this way it could be used from outside of the filters as well in the future. WIth the current implementation trying to reference it from the main library would cause a cyclic dependency import.

delic: Did things.

joohoi:
LGTM, thanks!

PR #56, opened 2019-08-06
Custom autocalibration strings

delic:
Added support for using multiple user-supplied auto-calibration strings. Each string is used in place of curent random strings, creating four calibration requests per string. Command line flag -acc, implies -ac.

Like so:
ffuf -w wordlist.txt -u http://www.example.com/FUZZ -acc foobar -acc xyzzy


joohoi:
LGTM! Could you add an changelog entry before we merge?

joohoi:
LGTM!

PR #59, opened 2019-08-21
Added redirect location in stdout output

SakiiR:


pkg/output/stdout.go
```diff
@@ -155,13 +155,28 @@ func (s *Stdoutput) resultQuiet(resp ffuf.Response) {
...
+	return redirectLocation
+}
+
 func (s *Stdoutput) resultNormal(resp ffuf.Response) {
 	var res_str string
+
+	redirectLocation := s.getRedirectLocation(&resp)
+	if redirectLocation != "" {
+		redirectLocation = fmt.Sprintf(", Redirect: %s", redirectLocation)
+	}
```
joohoi: I would prefer seeing this logic wrapped into a function alike `colorizeStatus` as it would keep the conditional logic outside of the `resultNormal` function here.

main.go
```diff
@@ -82,6 +82,7 @@ func main() {
 	flag.StringVar(&conf.Method, "X", "GET", "HTTP method to use")
 	flag.StringVar(&conf.OutputFile, "o", "", "Write output to file")
 	flag.StringVar(&opts.outputFormat, "of", "json", "Output file format. Available formats: json, csv, ecsv")
+	flag.BoolVar(&conf.ShowRedirectLocation, "l", false, "Show redirects response locations")
```
joohoi:
```suggestion
	flag.BoolVar(&conf.ShowRedirectLocation, "l", false, "Show target location of redirect responses")
```

pkg/output/stdout.go
```diff
@@ -155,13 +155,28 @@ func (s *Stdoutput) resultQuiet(resp ffuf.Response) {
 	}
 }
 
+func (s *Stdoutput) getRedirectLocation(resp *ffuf.Response) string {
```
joohoi: I'm thinking that a public method in [ffuf.Response](https://github.com/ffuf/ffuf/blob/master/pkg/ffuf/response.go) struct might be the best place for this, as then the code can be reused throughout the codebase.

SakiiR:
Is that fine now ?

pkg/output/stdout.go
SakiiR: ow Okay, did not see this one !

SakiiR:
Ready to merge if you may

joohoi:
This looks good and is ready to merge. Could you add a Changelog entry to https://github.com/ffuf/ffuf#changelog before that though?

pkg/output/stdout.go
```diff
@@ -156,14 +156,25 @@ func (s *Stdoutput) resultQuiet(resp ffuf.Response) {
...
-	fmt.Println(res_str)
+	fmt.Println(responseString)
+}
+
+// addRedirectLocation returns a formatted string containing the Redirect location or returns an empty string
+func (s *Stdoutput) addRedirectLocation(resp ffuf.Response) string {
+	if s.config.ShowRedirectLocation == true {
+		redirectLocation := resp.GetRedirectLocation()
+		if redirectLocation != "" {
+			return fmt.Sprintf(", Redirect: %s", redirectLocation)
```
joohoi: We're super constrained on the screen estate, especially when the multiple wordlist support lands. Because of this, I would prefer changing the word `Redirect: ` to just simply `->`.

SakiiR: I agree !

PR #71, opened 2019-09-25
Added lines count to filter/matcher and stdout + csv + json

SakiiR:


pkg/filter/filter.go
```diff
@@ -70,16 +74,24 @@ func calibrateFilters(j *ffuf.Job, responses []ffuf.Response) {
...
+	lineCalib = ffuf.UniqStringSlice(lineCalib)
 
 	if len(sizeCalib) > 0 {
 		AddFilter(j.Config, "size", strings.Join(sizeCalib, ","))
 	}
 	if len(wordCalib) > 0 {
 		AddFilter(j.Config, "word", strings.Join(wordCalib, ","))
 	}
+	if len(lineCalib) > 0 {
+		AddFilter(j.Config, "line", strings.Join(lineCalib, ","))
```
joohoi: I think that the concept of response line count is too broad to add to the auto-calibration. There's a good chance that all the `404`'s have a single line response, while all the actual legit responses we'd like to catch have a single line too.

SakiiR: Do you want to totally remove the concept of autocalibration on the line count ? :)

SakiiR: I copied all the code for the words like an idiot without really thinking :)

Just to have this feature which is pretty much the same than the word count one

joohoi: Yeah, I would like the line count to be removed from the autocalibration. I think it will contribute to a whole lot of misses, as it would filter everything based on the line count regardless of the return code after that.

joohoi:
Also now that the HTML & MD output format PR landed, could you also add the line count to these outputs?

joohoi:
Also: I don't want to push you, but I'm planning to make a new release during the weekend, so it would be great to have this feature in by then. The released version will act as a release candidate for 1.0, basically freezing the output formats and such to give guarantees of the format for projects that might integrate with ffuf.

pkg/filter/filter.go
SakiiR: all right

SakiiR:
> Also now that the HTML & MD output format PR landed, could you also add the line count to these outputs?

Done :) 

> Also: I don't want to push you, but I'm planning to make a new release during the weekend, so it would be great to have this feature in by then. The released version will act as a release candidate for 1.0, basically freezing the output formats and such to give guarantees of the format for projects that might integrate with ffuf.

That will be done for sure :)

SakiiR:
Lines count is now displayed in the markdown+html output :)

I also removed to auto calibration settings for the line count

Issue #39, opened 2019-06-14
Unsolicited response received on idle HTTP channel starting with "HTTP/1.0 408 Request Time-out\n


aels:
Faced this error so many-many-many times during scans:
[...]

PR #74, opened 2019-10-02
Add option for -debug-log.

ccsplit:
Both of these options have to do with the logging surrounding issues
such as #39. Where in that issue the server was returning data after
the connection was closed. Therefore, I added two options one for
completely disabling all of the internal logging functionality aka
sending it to /dev/null. Another for writing the logging information
to a file so it can be retrieved later if need be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,8 @@ To define the test case for ffuf, use the keyword `FUZZ` anywhere in the URL (`-
     	Wordlist file path or - to read from standard input
   -x string
     	HTTP Proxy URL
+  -debug-log string
+      Write the debug logging information to the specified file.
 ```
 
 eg. `ffuf -u https://example.org/FUZZ -w /path/to/wordlist`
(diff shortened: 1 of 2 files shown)


joohoi:
Thanks for the PR!

I think that it's rarely significant to see these messages, so I would propose the following: 

Using only a single CLI flag: `--debug-log` or so, and defaulting to logging off. Only try to write the logfile if the variable is set.

ccsplit:
@joohoi this has been changed to now only have the `-debug-log <file>` flag and will by default disable logging, it will also disable logging when an error occurs with creating a handle for the specified `<file>`. @aels, are you able to either provide the site you were having the issue with for testing, or build and run my [fork](https://github.com/ccsplit/ffuf) against the site you have to test the functionality? The site I had this issue against is no longer active, therefore, I currently don't have a site having this functionality to test against.

aels:
> @joohoi 
sorry, I'll have no time to test it again, but my query is
`ffuf -w ~/www/tools/alexatop1m.csv -u http://FUZZ/wp-includes/wlwmanifest.xml -of csv -mc 420 -t 8 -timeout 3 -mr "WordPress" -o wordpress2.txt`
where wordlist is from http://s3.amazonaws.com/alexa-static/top-1m.csv.zip, where the only column with domains left.
Hope it will help you.

ccsplit:
Thanks @aels, I tested it and it looks to be working correctly, one of the hosts with this issue is: `interia.pl`, one thing that would be nice, but not sure how difficult it would be, is to change the logger when `-debug-log` is active to add the URL it was testing as the log doesn't help that much.

joohoi:
This looks good and basically ready to merge. Could you add a [changelog](https://github.com/ffuf/ffuf#changelog) entry for this?

joohoi:
Oh, and seems like there's an merge conflict because of some PRs that were just merged.

ccsplit:
Should be resolved, added the flag to the Readme and the changelog as well.

PR #79, opened 2019-10-13
Multiple wordlist support

joohoi:
This PR enables usage of multiple wordlists and custom keywords.
The usage example is:
```
ffuf -u https://example.com/FIRST/SECOND -w "FIRST:/path/to/wordlist" -w "SECOND:/path/to/second/wordlist"
```
Todo:
- Change the help text for -w
- Figure out an output format that works, the current one is ugly

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ ffuf --input-cmd 'cat $FFUF_NUM.txt' -H "Content-Type: application/json" -X POST
 To define the test case for ffuf, use the keyword `FUZZ` anywhere in the URL (`-u`), headers (`-H`), or POST data (`-d`).
 
 ```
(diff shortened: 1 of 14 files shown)


BonJarber:
Note the formatting in this usage example appears to be backwards. It should be ffuf -u https://example.com/FIRST/SECOND -w "/path/to/wordlist:FIRST" -w "/path/to/second/wordlist:SECOND"

joohoi:
Nice catch @BonJarber ! Thanks for bringing it up. That was due to last minute changes to the UX (in order to support tab completion of file paths for wordlists) and apparently my brain was still in the old mode. I'll fix it right away.

PR #81, opened 2019-10-20
Match both %ext% and %EXT% in DirSearch compatibility mode

oh6hay:


pkg/input/wordlist.go
```diff
@@ -98,6 +98,11 @@ func (w *WordlistInput) readFile(path string) error {
 					contnt := strings.Replace(reader.Text(), "%EXT%", ext, -1)
 					data = append(data, []byte(contnt))
 				}
+			} else if strings.Index(reader.Text(), "%ext%") != -1 {
+				for _, ext := range w.config.Extensions {
+					contnt := strings.Replace(reader.Text(), "%ext%", ext, -1)
```
joohoi: I would rather see the original replace action changed to case insensitive regex replace. This way it would match possible typos `%Ext%` etc. as well.

oh6hay: better now? I haven't really written any go before

pkg/input/wordlist.go
```diff
@@ -91,11 +91,13 @@ func (w *WordlistInput) readFile(path string) error {
 
 	var data [][]byte
 	reader := bufio.NewScanner(file)
+	re := regexp.MustCompile(`(?i)%ext%`)
```
joohoi: For readability, this should get moved to the `if w.config.DirSearchCompat`... block

oh6hay: I think outside the loop is better so the regex doesn't get created for every line in the input

joohoi: Absolutely, thanks for clarification.

PR #113, opened 2019-11-25
count also 429 response codes if -sa (stop on all error cases) is used

delic:


pkg/ffuf/job.go
```diff
@@ -267,6 +284,11 @@ func (j *Job) CheckStop() {
 			}
 
 		}
+		if j.Config.StopOnAll && (float64(j.Count429)/float64(j.Counter) > 0.95) {
```
joohoi: We could probably lower this threshold value a bit. If a service has rate limit of, let's say 1000 req / minute, the first 1000 would get through, and the errors would start accumulating after that. To hit this threshold in this scenario, there would have to be 20 001 429 responses after that.

delic: Lowered to 20%

joohoi:
LGTM!

PR #115, opened 2019-12-01
Add flag for ignoring comments in wordlist

seblw:
Not sure if this is already implemented somehow, but I think it's nice to have it. Implemented as a flag to not break any default behaviour.

### DIFF
--- a/pkg/input/wordlist.go
+++ b/pkg/input/wordlist.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/ffuf/ffuf/pkg/ffuf"
 )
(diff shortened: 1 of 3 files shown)


pkg/input/wordlist.go
```diff
@@ -117,13 +118,23 @@ func (w *WordlistInput) readFile(path string) error {
 					data = append(data, []byte(contnt))
 				}
 			} else {
-				data = append(data, []byte(reader.Text()))
+				text := reader.Text()
+
+				if w.config.IgnoreWordlistComments && strings.HasPrefix(text, "#") {
```
joohoi: If we want to remove all the comments for entries like:
```
# this will be removed
 # this had an extra space in the beginning, and won't
valid_entry # this is a comment for the valid entry
```

The current code will  only catch the first one.

joohoi:
Hi @seblw , are you interested to change this PR to address the comments above, or would you like me to take over?

seblw:
Hi @joohoi, unfortunately I'm kind of a busy on some private stuff still and didn't have time to introduce fixes. I'm fine with you taking it over :)

joohoi:
Closing as this was implemented in #138

PR #124, opened 2019-12-16
fuzzing input in result, resolves #76

### LINKED ISSUE #76: Add the "fuzz word present in response" value in the response

delic:


pkg/filter/regex.go
```diff
@@ -15,12 +15,12 @@ type RegexpFilter struct {
...
 	re, err := regexp.Compile(value)
 	if err != nil {
-		return &RegexpFilter{}, fmt.Errorf("Size filter or matcher (-fs / -ms): invalid value: %s", value)
+		return &RegexpFilter{}, fmt.Errorf("Regexp filter or matcher (-fr / -mr): invalid value: %s", value)
 	}
 	return &RegexpFilter{Value: re, valueRaw: value}, nil
 }
 
-func (f *RegexpFilter) Filter(response *ffuf.Response) (bool, error) {
+func (f *RegexpFilter) Filter(response *ffuf.Response, inputs map[string][]byte) (bool, error) {
```
joohoi: You can access the fuzz keywords through `response.Request.Input` and can remove the code passing the input data through the application.

pkg/filter/regex.go
```diff
@@ -15,12 +15,12 @@ type RegexpFilter struct {
 func NewRegexpFilter(value string) (ffuf.FilterProvider, error) {
 	re, err := regexp.Compile(value)
 	if err != nil {
-		return &RegexpFilter{}, fmt.Errorf("Size filter or matcher (-fs / -ms): invalid value: %s", value)
+		return &RegexpFilter{}, fmt.Errorf("Regexp filter or matcher (-fr / -mr): invalid value: %s", value)
```
joohoi: Nice catch!

main.go
```diff
@@ -69,7 +69,7 @@ func main() {
 	flag.StringVar(&opts.delay, "p", "", "Seconds of `delay` between requests, or a range of random delay. For example \"0.1\" or \"0.1-2.0\"")
 	flag.StringVar(&opts.filterStatus, "fc", "", "Filter HTTP status codes from response. Comma separated list of codes and ranges")
 	flag.StringVar(&opts.filterSize, "fs", "", "Filter HTTP response size. Comma separated list of sizes and ranges")
-	flag.StringVar(&opts.filterRegexp, "fr", "", "Filter regexp")
+	flag.StringVar(&opts.filterRegexp, "fr", "", "Filter regexp, also fuzz keyword allowed")
```
joohoi: I don't think this needs a mention in the help text, as I think the default assumption here is that everything works with the keywords ;)

README.md
```diff
@@ -146,7 +146,7 @@ Usage of ./ffuf:
   -mode string
     	Multi-wordlist operation mode. Available modes: clusterbomb, pitchfork (default "clusterbomb")
   -mr string
-    	Match regexp
+    	Match regexp, also fuzz keyword allowed
```
joohoi: I don't think this needs a mention in the help text, as I think the default assumption here is that everything works with the keywords ;)

main.go
```diff
@@ -86,7 +86,7 @@ func main() {
 	flag.Var(&opts.cookies, "cookie", "Cookie data (alias of -b)")
 	flag.StringVar(&opts.matcherStatus, "mc", "200,204,301,302,307,401,403", "Match HTTP status codes from respose, use \"all\" to match every response code.")
 	flag.StringVar(&opts.matcherSize, "ms", "", "Match HTTP response size")
-	flag.StringVar(&opts.matcherRegexp, "mr", "", "Match regexp")
+	flag.StringVar(&opts.matcherRegexp, "mr", "", "Match regexp, also fuzz keyword allowed")
```
joohoi: I don't think this needs a mention in the help text, as I think the default assumption here is that everything works with the keywords ;)

pkg/filter/regex.go
```diff
@@ -29,6 +29,14 @@ func (f *RegexpFilter) Filter(response *ffuf.Response) (bool, error) {
 	}
 	matchdata := []byte(matchheaders)
 	matchdata = append(matchdata, response.Data...)
+	// is the raw value a fuzzing keyword
+	if _, fuzzed := inputs[f.valueRaw]; fuzzed {
+		matched, err := regexp.Match(string(inputs[f.valueRaw]), matchdata)
```
joohoi: We need to use `regexp.QuoteMeta` for the input data in order to properly escape all the regex control characters as the input data might have them.

delic: Tried searching such a method with no luck, good to know!

pkg/filter/regex.go
delic: Ah, missed that, thx.

main.go
delic: Removed.

README.md
delic: Removed.

main.go
delic: Removed.

pkg/filter/regex.go
delic: It is now fixed.

delic:
Will implement the more complex matching.

delic:
Now for example -mr 'FUZZ *[a-f0-9]+ FUZ2 t' using multiple wordlists works.

PR #127, opened 2019-12-30
-maxtime cli flag to limit running time of ffuf. resolves #85

### LINKED ISSUE #85: max time feature
It would be nice to have an feature similar to curl's --max-time option to bail out if ffuf has been running for more than N seconds.

delic:


joohoi:
LGTM!

PR #134, opened 2020-01-06
remove -k flag support, convert to dummy flag

Shaked:
Hey,

This is my first time contributing to `ffuf`.

This PR should remove -k support and convert the flag itself to a dummy in order to have backwards compatibility. 

`ffuf` will automatically set the `InsecureSkipVerify` flag to `true` as seen bellow. 

Please let me know if I missed something and how I can make it better.

Thank you 
Shaked


joohoi:
LGTM!

Issue #136, opened 2020-01-14
Feature request: Parse raw request to ffuf options


joohoi:
CLI flags `-request` and `-request-proto` should be added to parse a raw request into ffuf options.

The `-request-proto` is used to define the protocol to use as the raw HTTP request itself does not include that information. The possible values being `http` and `https`, defaulting to `https`. This option should only be used in conjunction with `-request`.

Options to parse in from the raw request file:
- Method
- Host
- Headers
- Body (POST/PUT/etc. data)

The parsing should be handled prior to applying other command line parameters to the configuration structure to allow the user to override some of the values in the parsed request.

As an example, a raw request file:
```
GET /FUZZ HTTP/1.1
Host: example.org
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
DNT: 1
```

Would be parsed into `ffuf` configuration structure equivalent of:
```
ffuf -u https://example.org/FUZZ -H "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0" -H "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8" -H "Accept-Language: en-US,en;q=0.5" -H "DNT: 1"
```

With the addition, and the raw request example file above, this would result into a `ffuf` command:
```
ffuf -request rawrequest.txt -w wordlist.txt
```

Changing the protocol to `http` instead of the default `https` would be done by:
```
ffuf -request rawrequest.txt -request-proto http -w wordlist.txt
```

Overriding the host to connect, the URL component and the protocol (but not the `Host` header!) could be handled by:
```
ffuf -request rawrequest.txt -u https://ffuf.io.fi/FUZZ -w wordlist.txt
````

And overriding only the `Host` header could be handled by:
```
ffuf -request rawrequest.txt -H "Host: somethingelse.tld" -w wordlist.txt
```

joohoi:
This issue has a [$100 bounty](https://twitter.com/codingo_/status/1217011896194199552) attached to it.

phspade:
This is super cool!

Ice3man543:
Hey, I tried to solve this in pr #137. Let me know if that's what we want.

[...]

PR #137, opened 2020-01-14
Added raw request parsing option 

Ice3man543:
Added feature to read and parse a request from the request body solving  https://github.com/ffuf/ffuf/issues/136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - master
   - New
+    - New CLI flag `-ic` to ignore comments from wordlist.
+    - New CLI flags `-request` to specify the raw request file to build the actual request from and `-request-proto` to define the new request format.
     - New CLI flag `-od` (output directory) to enable writing requests and responses for matched results to a file for postprocessing or debugging purposes.
     - New CLI flag `-maxtime` to limit the running time of ffuf
     - New CLI flags `-recursion` and `-recursion-depth` to control recursive ffuf jobs if directories are found. This requires the `-u` to end with FUZZ keyword.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 - master
   - New
-    - New CLI flag `-ic` to ignore comments from wordlist.
     - New CLI flags `-request` to specify the raw request file to build the actual request from and `-request-proto` to define the new request format.
     - New CLI flag `-od` (output directory) to enable writing requests and responses for matched results to a file for postprocessing or debugging purposes.
     - New CLI flag `-maxtime` to limit the running time of ffuf
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,4 +15,4 @@
 * [seblw](https://github.com/seblw)
 * [SolomonSklash](https://github.com/SolomonSklash)
 * [Shaked](https://github.com/Shaked)
-* [Ice3man543])(https://github.com/Ice3man543)
+* [Ice3man543](https://github.com/Ice3man543)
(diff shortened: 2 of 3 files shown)


joohoi:
Thanks for the PR, that was fast! The implementation looks good in general, there's just a few protocol corner cases that I would like to see handled. 

Would you think using [http.ReadRequest](https://golang.org/pkg/net/http/#ReadRequest) for the actual parsing part being a good fit? It addresses [some of the weirder parts](https://golang.org/src/net/http/request.go?s=33158:33209#L1064) automatically.

Could you also add a CHANGELOG.md entry as well as yourself to the CONTRIBUTORS.md?

Ice3man543:
Hey @joohoi, i did see that function but decided against it since i think my implementation covers all the basic use cases. I could implement this with `http.ReadRequest` but i think that will only introduce additional complexity. The current version is really simple and it just works. Let me know if you want me to add that?

joohoi:
Yeah, I'm fine with the current implementation in general. I'd like to see the one corner case I linked above to be handled though. Even if it's definitely a rarer format, it's still a valid request that I think `ffuf` should be able to parse correctly.

CONTRIBUTORS.md
```diff
@@ -15,3 +15,4 @@
 * [seblw](https://github.com/seblw)
 * [SolomonSklash](https://github.com/SolomonSklash)
 * [Shaked](https://github.com/Shaked)
+* [Ice3man543])(https://github.com/Ice3man543)
```
supertassu: remove extra  `)`

Ice3man543:
@joohoi, that's a fair point. I'll add that feature now. Also, can you check my other PR #138.

CONTRIBUTORS.md
Ice3man543: Nice catch, removing it now.

Ice3man543:
@joohoi, i just fixed the edge case issue and it works. Let me know what you think!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - master
   - New
+    - New CLI flag `-ic` to ignore comments from wordlist.
```
joohoi: This should be removed from here and added in #138

main.go
```diff
@@ -385,6 +400,66 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
...
+		if len(p) != 2 {
+			continue
+		}
+
+		conf.Headers[strings.TrimSpace(p[0])] = strings.TrimSpace(p[1])
+	}
+
+	// Handle case with the full http url in path. In that case,
+	// ignore any host header that we encounter and use the path as request URL
+	if !strings.HasPrefix(parts[1], "/") && strings.HasPrefix(parts[1], "http") {
```
joohoi:
```suggestion
	if strings.HasPrefix(parts[1], "http") {
```

joohoi: It can't be both ;)

main.go
```diff
@@ -385,6 +400,66 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
...
+
+		if err != nil || line == "" {
+			break
+		}
+
+		p := strings.SplitN(line, ":", 2)
+		if len(p) != 2 {
+			continue
+		}
+
```
joohoi: A minor UX detail here: ffuf recalculates `Content-Length` - header based on the actual request size regardless if a `Content-Length` header is provided or not. To avoid confusion from the printed banner that shows the parsed `Content-Length` header, I think we should skip adding it in the first place.

Ice3man543: Hey, this should be fixed in the latest commits/

CHANGELOG.md
Ice3man543: Fixed

joohoi:
LGTM!

PR #138, opened 2020-01-14
Added ignore comment option 

Ice3man543:
Ignore comments in multiple stuff - 
1. Ignore comment starting with comments.
2. Ignore comments in lines with space in front 
3. Strip comments from the end of the wordlist 

This addresses shortcomings that PR #115 failed to address.

### DIFF
--- a/pkg/input/wordlist.go
+++ b/pkg/input/wordlist.go
@@ -107,6 +107,7 @@ func (w *WordlistInput) readFile(path string) error {
 	defer file.Close()
 
 	var data [][]byte
+	var ok bool
 	reader := bufio.NewScanner(file)
 	re := regexp.MustCompile(`(?i)%ext%`)
 	for reader.Scan() {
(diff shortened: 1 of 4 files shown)


pkg/input/wordlist.go
```diff
@@ -117,17 +118,44 @@ func (w *WordlistInput) readFile(path string) error {
...
 	w.data = data
 	return reader.Err()
 }
+
+// stripComments removes all kind of comments from the word
+func stripComments(text string) string {
+	// If the line starts with a # ignoring any space on the left,
+	// return blank.
+	if strings.HasPrefix(strings.TrimLeft(text, " "), "#") {
+		return ""
```
joohoi: Lines that start with comment should be ignored completely instead of adding an empty entry, while still letting the user to actually include an empty entry to the wordlist.

To handle this the `stripComments` should return `(string, error)` or `(string, bool)` to figure out if we should skip the entry in the caller function.

Ice3man543: Fixed.

joohoi:
LGTM!

PR #169, opened 2020-02-17
Taking the Header User-Agent from a list

bjhulst:
Hi,
After attending the excellent workshop last weekend I found this feature also useful so I added it in my forked version of your codebase. 
Do you think this could be merged into the master release?
Regards,
Bob

### DIFF
--- a/help.go
+++ b/help.go
@@ -148,7 +148,7 @@ func Usage() {
 	fmt.Printf("    ffuf -w params.txt:PARAM -w values.txt:VAL -u https://example.org/?PARAM=VAL -mr \"VAL\" -c\n\n")
 
 	fmt.Printf("  Fuzz, alternate UserAgent-header, match HTTP 200 responses.\n")
-        fmt.Printf("    ffuf -w hosts.txt -A -u https://example.org/FUZZ  -mc 200\n\n")
+        fmt.Printf("    ffuf -w hosts.txt -A useragents.txt u https://example.org/FUZZ  -mc 200\n\n")
 
 
 	fmt.Printf("  More information and examples: https://github.com/ffuf/ffuf\n\n")
--- a/help.go
+++ b/help.go
@@ -61,7 +61,7 @@ func Usage() {
 		Description:   "",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
-		ExpectedFlags: []string{"ac", "acc", "c", "maxtime", "p", "s", "sa", "se", "sf", "t", "v", "V"},
+		ExpectedFlags: []string{"ac", "acc", "c", "maxtime", "p", "s", "sa", "se", "sf", "t", "v", "V", "A"},
 	}
 	u_compat := UsageSection{
 		Name:          "COMPATIBILITY OPTIONS",
@@ -146,6 +146,10 @@ func Usage() {
 
 	fmt.Printf("  Fuzz multiple locations. Match only responses reflecting the value of \"VAL\" keyword. Colored.\n")
 	fmt.Printf("    ffuf -w params.txt:PARAM -w values.txt:VAL -u https://example.org/?PARAM=VAL -mr \"VAL\" -c\n\n")
+
+	fmt.Printf("  Fuzz, alternate UserAgent-header, match HTTP 200 responses.\n")
+        fmt.Printf("    ffuf -w hosts.txt -A -u https://example.org/FUZZ  -mc 200\n\n")
+
 
 	fmt.Printf("  More information and examples: https://github.com/ffuf/ffuf\n\n")
 }
--- a/main.go
+++ b/main.go
@@ -22,7 +22,7 @@ import (
 type cliOptions struct {
 	extensions             string
 	delay                  string
-	filterStatus           string
+        filterStatus           string
 	filterSize             string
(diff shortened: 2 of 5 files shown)


joohoi:
Cheers! I'm really glad that you enjoyed the workshop. Also thanks for going through the trouble of familiarizing yourself with `ffuf` codebase and going through the trouble of writing this feature. However I wish to keep `ffuf` as bare bones as possible and to especially limit the number of command line flags to avoid confusion and to keep the tool approachable and for this reason would reject this PR.

My intent was to write here that you are already able to do that by:
```
ffuf -w hosts.txt -w useragents.txt:USERAGENT -u https://example.org/FUZZ  -H "User-agent: USERAGENT" -mc 200
```
which should work. I however realized that the current codebase has an bug with handling `User-agent` header; it can be replaced only if writing the header name in exactly the same capitalization than in the code. Because of this issue, the above command would not work while the following would do what it's supposed to:
```
ffuf -w hosts.txt -w useragents.txt:USERAGENT -u https://example.org/FUZZ  -H "User-Agent: USERAGENT" -mc 200
```

If you want to use a single, static UA from file, you can already do:
```
ffuf -w hosts.txt -u https://example.org/FUZZ  -H "User-Agent: `cat useragent.txt`" -mc 200
```

Would you be interested in fixing the bug with `User-Agent` header case sensitivity issue when using user defined value instead?

bjhulst:
Ok, thanks for the explanation.
If there is a bugreport for it already you can assign it to me. No problem. :-)

joohoi:
I created issue #171 to track this. However I'm unable to properly assign you through GitHub UI, but I'm going to make a comment noting this.

d3mondev:
Hello!

I would like to make an argument in favor of such a feature, as I was about to implement something similar myself.

The proposed workaround is to use multiple wordlists like so:

`ffuf -w words.txt:WORD -w useragents.txt:UA -u http://example.com/WORD -H "User-Agent: UA"`

This however has multiple drawbacks:

#### a) The default operation mode is "clusterbomb", meaning the addition of a wordlist increases the number of iterations exponentially

For a 100-word wordlist and a 200 lines useragents.txt file, I would expect to have 100 iterations, once for each wordlist. Instead, we get 100 * 200 iterations, resulting in many unnecessary checks as each word is tested against each useragent.

This also prevents the use of this workaround when performing any sort of fuzzing with multiple wordlists in clusterbomb mode.

```
        /'___\  /'___\           /'___\       
       /\ \__/ /\ \__/  __  __  /\ \__/       
       \ \ ,__\\ \ ,__\/\ \/\ \ \ \ ,__\      
        \ \ \_/ \ \ \_/\ \ \_\ \ \ \ \_/      
         \ \_\   \ \_\  \ \____/  \ \_\       
          \/_/    \/_/   \/___/    \/_/       

       v1.1.0-git
________________________________________________

 :: Method           : GET
 :: URL              : http://example.com/WORD
 :: Wordlist         : WORD: words.txt
 :: Wordlist         : UA: useragents.txt
 :: Header           : User-Agent: UA
 :: Follow redirects : false
 :: Calibration      : true
 :: Timeout          : 10
 :: Threads          : 40
 :: Matcher          : Response status: 200,204,301,302,307,401,403
________________________________________________

[...]

:: Progress: [20000/20000] :: Job [1/1] :: 10000 req/sec :: Duration: [0:00:02] :: Errors: 0 ::
```

#### b) Changing to "pitchfork" mode works better, but produces unnecessary iterations if the useragents wordlist is bigger than the fuzzing wordlist

For a 100-word wordlist and a 200 lines useragents.txt file, I would expect 100 iterations. Instead, ffuf will do 200 iterations and loop around the fuzzing wordlist unnecessarily because of pitchfork mode, duplicating the results in its output.

```
        /'___\  /'___\           /'___\       
       /\ \__/ /\ \__/  __  __  /\ \__/       
       \ \ ,__\\ \ ,__\/\ \/\ \ \ \ ,__\      
        \ \ \_/ \ \ \_/\ \ \_\ \ \ \ \_/      
         \ \_\   \ \_\  \ \____/  \ \_\       
          \/_/    \/_/   \/___/    \/_/       

       v1.1.0-git
________________________________________________

 :: Method           : GET
 :: URL              : http://example.com/WORD
 :: Wordlist         : WORD: words.txt
 :: Wordlist         : UA: useragents.txt
 :: Header           : User-Agent: UA
 :: Follow redirects : false
 :: Calibration      : false
 :: Timeout          : 10
 :: Threads          : 40
 :: Matcher          : Response status: 200,204,301,302,307,401,403
________________________________________________

[...]

:: Progress: [200/200] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
```

As I find myself testing against a WAF that can easily be bypassed by specifying a random user agent, I think the User-Agent header should be treated as an edge case and that this feature should be allowed.

If this is ever implemented, I would actually prefer to:

* Use -ua as a command line argument, as it would be more in line with the naming convention of the other arguments already present in ffuf (-rua for "random user agent" would perhaps be even more accurate?)
* Specify the filename myself

Typical usage would simply become:

`ffuf -w words.txt -ua useragents.txt -u http://example.com/FUZZ`

Thanks,
d3mondev

Issue #171, opened 2020-02-18
Make defining User-agent header case insensitive


joohoi:
The `User-agent` header is the only one ffuf sets up automatically. There's currently a bug that makes defining a custom user agent header name case sensitive;

This does not work:
```
ffuf -H "User-agent: custom string" ...
```

While this does:
```
ffuf -H "User-Agent: custom string" ...
```

joohoi:
As noted in #169 @bjhulst wishes to tackle this issue.

bjhulst:
Roger.

joohoi:
This was fixed in #173 

Thanks again @bjhulst !

okondo-afo:
FFUF returns incorrect data with its useragent header, all ffuf data is wrong , it kept returning 403 forbidden when its not

spidy-idc:
> Project37 reacted with eyes emoji

Some websites block ffuf and gobuster user agents, We have to randomize user agents for fuzzing

netwons:
how randomize user agents?

PR #173, opened 2020-02-18
 Make defining User-agent header case insensitive #171 

bjhulst:
This small PR only takes the value as mentioned in the bug report as other possible option.

I did not compare or sanitize the whole header nor looked at duplicate Header values for User-Agent.
Btw. The IETF RFC for HTTP Header has only defined:  "User-Agent:   custom string"


pkg/runner/simple.go
```diff
@@ -93,11 +93,17 @@ func (r *SimpleRunner) Execute(req *ffuf.Request) (ffuf.Response, error) {
...
-		req.Headers["User-Agent"] = fmt.Sprintf("%s v%s", "Fuzz Faster U Fool", ffuf.VERSION)
+		if _, ok := req.Headers["User-agent"]; !ok {
+			req.Headers["User-Agent"] = fmt.Sprintf("%s v%s", "Fuzz Faster U Fool", ffuf.VERSION)
+		} else { 
+			req.Headers["User-Agent"] = req.Headers["User-agent"]
+		}
 	}
-	// Handle Go http.Request special cases
+	
+        // Handle Go http.Request special cases
```
joohoi: These whitespace and comment changes should be reverted

main.go
```diff
@@ -401,7 +401,7 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
 		}
 	}
 
-	// Auto-calibration strings
+        // Auto-calibration strings
```
joohoi: This whitespace change should be reverted.

pkg/runner/simple.go
```diff
@@ -93,11 +93,17 @@ func (r *SimpleRunner) Execute(req *ffuf.Request) (ffuf.Response, error) {
 	if err != nil {
 		return ffuf.Response{}, err
 	}
+
 	// Add user agent string if not defined
 	if _, ok := req.Headers["User-Agent"]; !ok {
-		req.Headers["User-Agent"] = fmt.Sprintf("%s v%s", "Fuzz Faster U Fool", ffuf.VERSION)
+		if _, ok := req.Headers["User-agent"]; !ok {
```
joohoi: HTTP header names are case insensitive as defined in [RFC 2616 section 4.2](https://ietf.org/rfc/rfc2616.html#section-4.2)
Because of this we should account for all possible names, including `uSEr-agEnt` etc. 

The proper way to handle it would be iterating through `req.Headers` keys, and check if any of them contain `user-agent` by using case-insensitive matching provided by [strings.EqualFold()](https://golang.org/pkg/strings/#EqualFold)

pkg/runner/simple.go
bjhulst: reverted

main.go
bjhulst: reverted

bjhulst:
I updated the code and noticed it would be better to move the default User-Agent header parsing to the main.go part. It needs to be done only once for every program start and not in every jobrunner request. I also used the CanonicalMIMEHeaderKey to make them more generic. This also reduced the need to check for duplicate -H values.

main.go
```diff
@@ -335,15 +336,24 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
...
 	}
 
-	//Prepare headers
+	//Prepare headers and make canonical
 	for _, v := range parseOpts.headers {
 		hs := strings.SplitN(v, ":", 2)
 		if len(hs) == 2 {
-			conf.Headers[strings.TrimSpace(hs[0])] = strings.TrimSpace(hs[1])
+			// trim and make canonical	
+			var CanonicalHeader string = textproto.CanonicalMIMEHeaderKey (strings.TrimSpace(hs[0]))
```
joohoi: Coding style things like whitespace on this line and the alphabetical import order get automatically fixed when you run `go fmt` in the project directory. We should do that before merging the changes in master.

bjhulst:
I updated the code and docs.

pkg/runner/simple.go
```diff
@@ -73,7 +73,8 @@ func (r *SimpleRunner) Prepare(input map[string][]byte) (ffuf.Request, error) {
 		req.Method = strings.Replace(req.Method, keyword, string(inputitem), -1)
 		headers := make(map[string]string, 0)
 		for h, v := range req.Headers {
-			headers[strings.Replace(h, keyword, string(inputitem), -1)] = strings.Replace(v, keyword, string(inputitem), -1)
+			var CanonicalHeader string = textproto.CanonicalMIMEHeaderKey (strings.Replace(h, keyword, string(inputitem), -1))
```
joohoi: This still requires the `go fmt` treatment

bjhulst: done.

bjhulst:
> This will end up in the headers slice as "Fuzz : customvalue" and as such, will not be replaced by the input wordlist values when creating the request. This makes me think if _still_ should do this heavy lifting and replacement in the job preparation phase... What do you think?

We could canonicalize all header keys except that contains "FUZZ:".

joohoi:
> We could canonicalize all header keys except that contains "FUZZ:".

This solution is good, but the issue with this is that the user is able to define their own keywords when using `-w` and `--input-cmd`. Those are stored in `conf.CommandKeywords` though, so we should look through that list and figure out if one is defined as header name and skip canonicalization if so.

bjhulst:
So skip canonical if key = FUZZ or key in conf.CommandKeywords   ?

joohoi:
> So skip canonical if key = FUZZ or key in conf.CommandKeywords ?

Yeah, basically this. However "FUZZ" is automatically added to `conf.CommandKeywords` if it should be used. So looping through the `conf.CommandKeywords` should be sufficient.

An added benefit of your approach is, by the way giving the user more transparency over the request being sent by showing the default "User-Agent" string in the config banner displayed in the beginning of each run.

bjhulst:
Looping now both CommandKeywords and InputProviders to make sure they are not canonical afterwards.

main.go
```diff
@@ -335,15 +336,42 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
...
 		} else {
 			errs.Add(fmt.Errorf("Header defined by -H needs to have a value. \":\" should be used as a separator"))
 		}
 	}
+
+	// set default User-Agent header if not present
+	if _, ok := conf.Headers["User-Agent"]; !ok {
+		conf.Headers["User-Agent"] = fmt.Sprintf("%s v%s", "Fuzz Faster U Fool", ffuf.VERSION)
+	}
+
```
joohoi: Unfortunately I think we need to move this part of the logic back to the `simple.go`:

If one of the headers is `-H "FUZZ: somevalue"`, this gets added to the list of headers as there's no header named `User-Agent` present, only the keyword. In go, maps are not ordered so there's a considerable chance that `FUZZ: somevalue` gets replaced and added to the result map _before_ we iterate over this native `User-Agent`. The `User-Agent: Fuzz Faster U Fool ...` will then replace the `User-Agent: somevalue` we added to the result map before.

If we do conditional statements in the `main.go` to figure out if we should add it in the first place, we still would need to do some checks and additions in `simple.go`. Because of this, I think the right call is to move this back to `simple.go`, after the canonicalization and replacing the keyword values. It should work as it is though.

bjhulst: No problem. This is still far less complex than the Go language client for Ethereum. ;-)
I will move the codeblock back to simple.go.

Maybe another PR could be to build unit-test cases for all possible combinations.

bjhulst:
The 2 conflicting files are now only text conflicts, not code. (still need to move the logic to simple.go...)

Issue #179, opened 2020-02-24
Make -maxtime count towards total execution time regardless of new jobs added by -recursion


joohoi:
The `--maxtime` CLI flag should control the total execution time. However, the timer is reset for each new job added to the list by recursion detection that gets enabled by `--recursion`.

Additionally, add a new CLI flag `--maxtime-job` that counts towards the maximum execution time for each job, and if it hits, cancel the job and move forward to the next one.

Damian89:
Thanks for adding the issue.

OriBoharon:
Is this still open after #180 ?

PR #180, opened 2020-02-25
Fix for issue #179

Damian89:
__Usage:__

This will stop entire execution of ffuf after 15 seconds (it of course waits till requests got their responses):

`-maxtime 15 `

This will stop the current running job after 5 seconds and continue with the next one:

`-maxtime-job 5`

You can combine both. This will allow max exec. time of 60 seconds but each job (even the one on level 0) is allowed to run 20 seconds.

`-maxtime 60 -maxtime-job 20`

Those settings are especially useful when working in recursion mode+big wordlist+a lot of new jobs are added by ffuf because there are a lot of folders.


pkg/ffuf/job.go
```diff
@@ -367,19 +394,36 @@ func (j *Job) CheckStop() {
...
 			j.Stop()
 		}
 	}
+
+	// Check for runtime of current job
+	if j.Config.MaxTimeJob > 0 {
+		dur := time.Now().Sub(j.startTimeJob)
+		runningSecs := int(dur / time.Second)
+		if runningSecs >= j.Config.MaxTimeJob {
+			j.Output.Info(fmt.Sprintf("Maximum running time for this job reached, continuing with next job if one exists."))
```
joohoi: I think this should not directly output the info, but instead set the `j.Error` like the full process stop case does.

```suggestion
			j.Error = "Maximum running time for this job reached, continuing with next job if one exists."
```

pkg/ffuf/progress.go
```diff
@@ -5,10 +5,11 @@ import (
...
 
 type Progress struct {
-	StartedAt  time.Time
-	ReqCount   int
-	ReqTotal   int
-	QueuePos   int
-	QueueTotal int
-	ErrorCount int
+	StartedAt        time.Time
+	StartedAtProcess time.Time
```
joohoi: This isn't currently used anywhere in the progress output, so I think it should be removed.

pkg/ffuf/job.go
```diff
@@ -175,25 +194,33 @@ func (j *Job) runProgress(wg *sync.WaitGroup) {
...
 func (j *Job) updateProgress() {
 	prog := Progress{
-		StartedAt:  j.startTime,
-		ReqCount:   j.Counter,
-		ReqTotal:   j.Input.Total(),
-		QueuePos:   j.queuepos,
-		QueueTotal: len(j.queuejobs),
-		ErrorCount: j.ErrorCounter,
+		StartedAt:        j.startTimeJob,
+		StartedAtProcess: j.startTime,
```
joohoi: This isn't currently used anywhere in the progress output, so I think it should be removed.

pkg/ffuf/progress.go
Damian89: StartedAt is used in stdout.go to calculate the duration per job, so only removed StartedAtProcess.

pkg/ffuf/job.go
Damian89: Same here, removed StartedAtProcess.

pkg/ffuf/job.go
Damian89: Yep, noticed that as well. Fixed.

README.md
```diff
@@ -63,6 +63,22 @@ This is a very straightforward operation, again by using the `FUZZ` keyword. Thi
...
+If you don't want ffuf to run indefinitely, you can use the `-maxtime`. This stops __the entire__ process after a given time (in seconds).
+
+```
+ffuf -w /path/to/wordlist -u https://target/FUZZ -maxtime 60
+```
+
+When working with recursion, you can control the maxtime __per job__ using `-maxtime-job`. This will stop the current job after a given time (in seconds) and continue with the next one.
+
+```
+ffuf -w /path/to/wordlist -u https://target/FUZZ -maxtime-job 60
```
joohoi: I think it would be good to add `-recursion` to example CLI, as well as describe what a "sub" job is in the description to clarify the inner workings.

README.md
```diff
@@ -71,10 +71,10 @@ If you don't want ffuf to run indefinitely, you can use the `-maxtime`. This sto
 ffuf -w /path/to/wordlist -u https://target/FUZZ -maxtime 60
 ```
 
-When working with recursion, you can control the maxtime __per job__ using `-maxtime-job`. This will stop the current job after a given time (in seconds) and continue with the next one.
+Especially when working with recursion, you can control the maxtime __per job__ using `-maxtime-job`. This will stop the current job after a given time (in seconds) and continue with the next one. Internally a job is the brute forcing of one specific folder. This is also the case for the root folder.
```
joohoi:
```suggestion
When working with recursion, you can control the maxtime __per job__ using `-maxtime-job`. This will stop the current job after a given time (in seconds) and continue with the next one. New jobs are created when the recursion functionality detects a subdirectory. 
```

CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
 
 - master
   - New
-  - Changed
```
joohoi: This line was apparently deleted by mistake.

CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
...
+    - New CLI flag `-maxtime-job` to set max. execution time per job.
+    - Changed behaviour of `-maxtime`, can now be used for entire process.
 
 - v1.0.2
-  - Changed
```
joohoi: This line was apparently deleted by mistake.

PR #206, opened 2020-04-06
Add Output Checkpoint option

### LINKED ISSUE #205: ffuf should write output while a job is in progress
When doing a long-running `ffuf` it would be useful to have the output file (e.g. JSON) be written periodically, based on results seen so far, so that a user can get a head-start on exploring the results (Preventing https://xkcd.com/303/ :wink:)

justinsteven:
The option specifies how often output writing should be done (every n
requests) while a job is in progress

This is a naive attempt to fix #205, I'm new to Go and `ffuf` - guidance would be appreciated!

Known issues:

* File writing isn't atomic (But then again it never was). Repeatedly reading the Output File outside of `ffuf` doesn't guarantee you're getting complete and hence legal output
* It doesn't care for concurrency? A low enough `-oc` value might result in parallel attempts to write to the Output File?
* I don't really know how jobs work, or if this could cause issues with recursive mode (Something I don't use)
* I don't know if we can trust `j.Counter` for this kind of thing. I've sometimes seen the `func()` loop inside of `startExecution()` be executed with identical `j.Counter` values. This is especially evident at the end of `ffuf`'s execution, perhaps when we're waiting for the final requests to come back? If we're unlucky, and at that time `j.Counter` is a multiple of `-oc`, we'll be writing the Output File (In parallel threads?) in a tight loop?
* Provides a potentially needless config knob. If we could fix the above issues, perhaps we could just always checkpoint the output with a sane `-oc` value?

### DIFF
--- a/help.go
+++ b/help.go
@@ -96,7 +96,7 @@ func Usage() {
 		Description:   "Options for output. Output file formats, file names and debug file locations.",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
-		ExpectedFlags: []string{"debug-log", "o", "of", "od"},
+		ExpectedFlags: []string{"debug-log", "o", "of", "od", "oc"},
 	}
 	sections := []UsageSection{u_http, u_general, u_compat, u_matcher, u_filter, u_input, u_output}
 
--- a/main.go
+++ b/main.go
@@ -103,6 +103,7 @@ func main() {
 	flag.StringVar(&conf.OutputFile, "o", "", "Write output to file")
 	flag.StringVar(&opts.outputFormat, "of", "json", "Output file format. Available formats: json, ejson, html, md, csv, ecsv")
 	flag.StringVar(&conf.OutputDirectory, "od", "", "Directory path to store matched results to.")
+	flag.IntVar(&conf.OutputCheckpoint, "oc", -1, "Write output to file every n requests (-1 to disable).")
 	flag.BoolVar(&conf.IgnoreBody, "ignore-body", false, "Do not fetch the response content.")
 	flag.BoolVar(&conf.Quiet, "s", false, "Do not print additional information (silent mode)")
 	flag.BoolVar(&conf.StopOn403, "sf", false, "Stop when > 95% of responses return 403 Forbidden")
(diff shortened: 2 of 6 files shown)


joohoi:
Thanks for the PR! I like the idea, and even more so: I think it should be made default.

However, I don't think we should rely on `j.Counter` for triggering the write event but instead do a write after every `n` seconds. I'm thinking of something like once every 3-5 minutes or so.

To combat  potential concurrency issues a new mutex lock could be used. However it's important to re-check the need to write within the function protected by the lock in order to not to end up in a situation where multiple coroutines think that they should write the output file, and are just merely waiting for the mutex unlock.

About the jobs, I think there might be a bug there. I'm going to review the output file writing in case of recursive functionality in general before tackling this PR (later today or tomorrow).

justinsteven:
Happy to reject this PR and take discussion to the issue

I like making it the default, and not needing to expose an "every n (minutes|requests)" knob to the user. But some people might want to get output more frequently than the default, or even on demand. What do you think of catching SIGUSR1 and using it as a signal to dump output (Unless output is currently being dumped, per the mutex)

Or would this be a pain cross-platform? (I know it couldn't be supported under Windows, and I don't know how portable SIGUSR1 is to *bsd and macOS)

joohoi:
Hooking to `SIGUSR1` sounds good to me. Definitely not a pain implementing it cross-platform, as we can just decide to not to try to scramble to support this in Windows.

To handle the different behavior depending on the platform we can leverage build tags to have different implementations of `func (j *Job) interruptMonitor()`  for windows and POSIX systems. Example of build tags in use can already be found in [const_windows.go](https://github.com/ffuf/ffuf/blob/master/pkg/output/const_windows.go) and [const.go](https://github.com/ffuf/ffuf/blob/master/pkg/output/const.go)

joohoi:
> About the jobs, I think there might be a bug there. I'm going to review the output file writing in case of recursive functionality in general before tackling this PR (later today or tomorrow).

This was good, and fortunately I had the wrong hunch on this one.

justinsteven:
@joohoi I'll leave this in your capable hands. I'm clearly out of my depth, my patch seems to be leaking memory :sweat_smile:

justinsteven:
Update: Maybe there’s no leak. Maybe I was just doing something silly.

justinsteven:
Closing this. The code is stale and bad, and I think the upcoming interactive mode solves most of the problem. Taking discussion to #205

PR #210, opened 2020-04-09
Added option to hide the banner

samirettali:


help.go
```diff
@@ -61,7 +61,7 @@ func Usage() {
 		Description:   "",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
-		ExpectedFlags: []string{"ac", "acc", "c", "maxtime", "maxtime-job", "p", "s", "sa", "se", "sf", "t", "v", "V"},
+		ExpectedFlags: []string{"ac", "acc", "c", "maxtime", "maxtime-job", "nb", "p", "s", "sa", "se", "sf", "t", "v", "V"},
```
joohoi, line 64: In case we go with writing to `stderr`, we don't need to add a new CLI flag.

main.go
```diff
@@ -104,6 +104,7 @@ func main() {
 	flag.StringVar(&opts.outputFormat, "of", "json", "Output file format. Available formats: json, ejson, html, md, csv, ecsv")
 	flag.StringVar(&conf.OutputDirectory, "od", "", "Directory path to store matched results to.")
 	flag.BoolVar(&conf.IgnoreBody, "ignore-body", false, "Do not fetch the response content.")
+	flag.BoolVar(&conf.NoBanner, "nb", false, "Do not print the banner ")
```
joohoi, line 107: In case we go with writing to `stderr`, we don't need to add a new CLI flag.

pkg/ffuf/config.go
```diff
@@ -21,6 +21,7 @@ type Config struct {
 	OutputFile             string                    `json:"outputfile"`
 	OutputFormat           string                    `json:"outputformat"`
 	IgnoreBody             bool                      `json:"ignorebody"`
+	NoBanner               bool                      `json:"nobanner"`
```
joohoi, line 24: In case we go with writing to `stderr`, we don't need to add a new configuration variable.

samirettali:
Hi, sorry for the delay, I've been quite busy lately.
The reason I added the option is because when writing writeups from CTFs I almost always remove the banner because it's too big and it makes the great majority of the output.

joohoi:
Ok, that sounds reasonable. I'd still like to output the banner to Stderr, and I think it'd be a good match to make that change in context of this PR. Could you additionally also change the usage text on `README.md` to match the changes. Oh and add yourself to the alphabetically correct place in `CONTRIBUTORS.md` :)

Issue #215, opened 2020-04-16
Support all output formats concurrently


Daviey:
It would be really useful to be able to output all the format types concurrently as the different representations can be useful later on (considering the amount of storage required is small).

Example workflow:

- HTML is useful for quickly reviewing in a GUI friendly way
- CSV is useful to import into Excel
- JSON is useful to import into into custom tooling to programmatically perform further investigation.
- MD useful to import into a markdown based report document.

Suggesting:
`-of all -o output/report`

Where a suffix would be added with {json,ejson,html,md,csv,ecsv} etc.

PR #218, opened 2020-04-16
Support outputting all file formats concurrently

Daviey:
Previously ffuf would allow outputting to a single
file-format, ie Markdown (md), json or HTML.  It was not
possible to output in multiple formats in the same
execution.

This change allows specifying an output (-of) of "all",
which means that the output filename (-o) is used, but the
appropriate suffix is added.

As an example,
... -of all -o output/report

Will output:
 - output/report.json
 - output/report.html
 - output/report.csv
 - ... etc

Fixes ffuf/ffuf#215

Signed-off-by: Dave Walker (Daviey) <email@daviey.com>

### DIFF
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -83,7 +83,16 @@ func (s *Stdoutput) Banner() error {
 
 	// Output file info
 	if len(s.config.OutputFile) > 0 {
-		printOption([]byte("Output file"), []byte(s.config.OutputFile))
+
+		// Use filename as specified by user
+		OutputFile := s.config.OutputFile
+
+		if s.config.OutputFormat == "all" {
+			// Actually... append all extensions
+			OutputFile += ".{json,ejson,html,md,csv,ecsv}"
+		}
+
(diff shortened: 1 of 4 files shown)


pkg/output/stdout.go
```diff
@@ -196,10 +205,59 @@ func (s *Stdoutput) Warning(warnstring string) {
...
+	}
+
+	s.config.OutputFile = BaseFilename + ".csv"
+	err = writeCSV(s.config, s.Results, false)
+	if err != nil {
+		s.Error(fmt.Sprintf("%s", err))
+	}
+
+	s.config.OutputFile = BaseFilename + ".csv"
+	err = writeCSV(s.config, s.Results, true)
```
joohoi, line 246: I believe this is supposed to call eCSV instead of CSV (also with a correct file extension)

Daviey: Good catch, the file extension was a mistake.  However, it is still the writeCSV function, but with true or false.  Fixing now!

PR #226, opened 2020-04-27
fix file get thing

### LINKED ISSUE #225: GET request supplied with --request sent as POST
Looks like it's due to these lines:
https://github.com/ffuf/ffuf/blob/master/main.go#L456-L460

as the data is always set when parsing the request file:
https://github.com/ffuf/ffuf/blob/master/main.go#L537

Should be a trivial fix, will PR in a few minutes

C-Sto:
Should resolve #225 

Very minimal changes, so I doubt it will introduce any issues, but for the record I didn't thoroughly verify it.

Not sure what you'd like to do with versioning etc :)

### DIFF
--- a/main.go
+++ b/main.go
@@ -453,10 +453,12 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
 	}
 
 	// Handle copy as curl situation where POST method is implied by --data flag. If method is set to anything but GET, NOOP
-	if conf.Method == "GET" && len(parseOpts.request) == 0 {
-		if len(conf.Data) > 0 {
-			conf.Method = "POST"
-		}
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -453,7 +453,7 @@ func prepareConfig(parseOpts *cliOptions, conf *ffuf.Config) error {
 	}
 
 	// Handle copy as curl situation where POST method is implied by --data flag. If method is set to anything but GET, NOOP
-	if conf.Method == "GET" {
+	if conf.Method == "GET" && len(parseOpts.request) == 0 {
```
joohoi, line 456: Can you add comment here to point out why this condition exists. It'll help figuring out what's going on in the future :)

Issue #232, opened 2020-05-03
Recursive doesn't work when URL contains port number


Daviey:
When a port is speciiced as part of the url, such as `-u http://example.com:80/FUZZ` and `-recursion`, then any found 'folders' are not added to the task as it should be (and is) if the port is omitted.

This is because handleRecursionJob() compares the request with an added trailing /, against a potential redirect and GetRedirectLocation() neglects to add the port number, which means the if statement is wrongly determining that it isn't a folder (as the if statement will always be false).

https://github.com/ffuf/ffuf/blob/08ec6bad2aad1e265b20d9036a16a8f30679ed3e/pkg/ffuf/job.go#L315-L318

This is effectively doing:
`if ("http://example.com:80/foobar" + "/") != "http://example.com/foobar/" {`

I took a quick look at trying to add correct handling to GetRedirectLocation(), but making it work with both URL's with port and without is a challenge.  Wondering if `--port` should be added as an explicit flag?

Issue #192, opened 2020-03-19
Cannot use follow_redirect and recursion at the same time


Kiblyn11:
Hello,

I would like to report an issue about options follow_redirect and recursion.

## Issue
Cannot use follow_redirect and recursion option at the same time.
If I use both, recursion doesnt work / trigger when a directory is found.

## Current behavior
Example command:
```
ffuf -w .../SecLists/Discovery/Web-Content/raft-medium-directories-lowercase.txt -u https://www.google.com/FUZZ -recursion-depth 4 -t 10 -fc 403 -recursion -r
```
No recursion occurs when a directory is found
If I doesnt specify `-r` option, recursion occurs, but it also occurs on 301 responses which lead to 404 sometimes.

## Expected behavior
I would like to be able to combine `-r` and `-recursion` option to avoid fuzzing 301 responses which lead to 404

joohoi:
Thanks for opening the issue!

This is kind of hard one to solve, as the concept of "directory" is just an URL that redirects to the same URL with appended `/`. And I don't think it'd be a good idea to implement "magic" that would handle this functionality per default as some configurations return `404` for everything but exact matches. ( https://target.tld/path/ - 404, https://target.tld/path/resource.ext - 200 )

As of now, I think your best bet might be scripting around ffuf to check the redirect results and create new runs based on this scripted filtering.

PR #233, opened 2020-05-03
Fix recursive when URL contains explicit port

Daviey:
Previously when handling recursive jobs, found folders were
not being added if the user supplied url (-u) contained a
port number (ie, http://example.com:80/FUZZ).

This is because the logic to determine if a redirect was a
folder was to compare if the redirected URL == the
request-url + "/".  (ie, is a trailing "/" added)

The issue is that GetRedirectLocation() strips the port,
which means the if statement is wrongly always false (as the
port is missing).

This change now checks if the redirection has a trailing
"/", which means we can imply the redirection is a folder,
as that is the real factor that needs testing.

The issue with this fix is that if a redirect goes to a
whole new url, then we have *potentially* implemented follow
redirects on recursive (ffuf/ffuf#192) which is probably not
desirable?

Fixes ffuf/ffuf#232

Signed-off-by: Dave Walker (Daviey) <email@daviey.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     - Added tls renegotiation flag to fix #193 in http.Client
     - Fixed HTML report to display select/combo-box for rows per page (and increased default from 10 to 250 rows).
     - Added Host information to JSON output file
+    - Fixed recursive not adding jobs on discovered folders when user supplied url (-u) contains a port number.
 
 - v1.0.2
   - Changed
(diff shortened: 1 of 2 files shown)


joohoi:
Sorry about the late answer. IIRC we discussed this out of band, and a more advanced logic is needed for figuring out recursion jobṣ. 

With this proposed change, a ffuf job can easily spiral out of control to out-of-scope domains and URLs. 

For example, if a response for https://example.org/something would contain `Location: https://google.com/` , a recursion job would be created for `https://google.com/FUZZ`

Issue #48, opened 2019-07-07
[Feature Request] URL Encoding Option


ngkogkos:
Hello,

It would be nice if ffuf would have an option to URL encode FUZZ input as many lists contain spaces or other special characters.

It could accept a list of characters to auto-URL encode for instance:

`-url-encode " #&" `

joohoi:
Hi! That's a good idea! Do you think there's value in defining the characters to encode instead of just going all-out and encoding all of them?

ngkogkos:
Hi, yes for sure there is value!

Some web apps decide to not URL decode stuff back some times which could lead into researchers missing XSS for instance. 

I like having the option to decide based on how the specific app works, but for some characters it really needs to be mandatory such as spaces # and & as they would break the GET requests otherwise.

joohoi:
That's a sound explanation, thanks for the input! Sure sounds like it would be beneficial to define a list. I'll probably add an `all` keyword to do the blanket - encode all.

ngkogkos:
Amazing thank you so much :D, I can't wait to test it!

Shaddy:
wfuzz has an entire set of modules that allows different encodings, really interesting because it lets you do double %25xx escapes, base64 encoding or even md5. 

```
Available encoders:

  Category      | Name              | Summary
------------------------------------------------------------------------------------------------------------------------
  hashes        | base64            | Encodes the given string using base64
  url           | doble_nibble_hex  | Replaces ALL characters in string using the %%dd%dd escape
  url_safe, url | double urlencode  | Applies a double encode to special characters in string using the %25xx escape.
                |                   | Letters, digits, and the characters '_.-' are never quoted.
  url           | first_nibble_hex  | Replaces ALL characters in string using the %%dd? escape
  default       | hexlify           | Every byte of data is converted into the corresponding 2-digit hex representatio
                |                   | n.
  html          | html_decimal      | Replaces ALL characters in string using the &#dd; escape
  html          | html_escape       | Convert the characters &<>" in string to HTML-safe sequences.
  html          | html_hexadecimal  | Replaces ALL characters in string using the &#xx; escape
  hashes        | md5               | Applies a md5 hash to the given string
  db            | mssql_char        | Converts ALL characters to MsSQL's char(xx)
  db            | mysql_char        | Converts ALL characters to MySQL's char(xx)
  default       | none              | Returns string without changes
  db            | oracle_char       | Converts ALL characters to Oracle's chr(xx)
[...]

PR #236, opened 2020-05-07
support for base64 and urlencoding

kazkansouh:
I often find myself needing to switch back to other tools as there is a lack of support for url encoding in ffuf. So I have put together this small PR in hopes its useful, and in line with what is discussed in #48.

* added two command line options `-el` encoding list, and `-ep` encoding parameters.
* the `-w` parameter now accepts encoders appended after the keyword (similar to `wfuzz`)

So now its possible to do something like the following and have the cookie automatically url encoded:

```
$ echo  '9lanBYqNH2wezQcQ5JiGl7070PzSIfSI' | ./bin/radamsa -m bf --count 100 | ffuf -u http://xx.xx.xx.xx/login.php -w -:FUZZ:urlenc -H 'Cookie: auth=FUZZ' -mr admin
```

I am sure this can easily be extended with other encoders if needed. Although at the moment I have just left it with the two I generally find useful.

### DIFF
--- a/help.go
+++ b/help.go
@@ -89,7 +89,7 @@ func Usage() {
 		Description:   "Options for input data for fuzzing. Wordlists and input generators.",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
-		ExpectedFlags: []string{"D", "ic", "input-cmd", "input-num", "mode", "request", "request-proto", "e", "w"},
+		ExpectedFlags: []string{"D", "ic", "input-cmd", "input-num", "mode", "request", "request-proto", "e", "w", "el", "ep"},
 	}
 	u_output := UsageSection{
 		Name:          "OUTPUT OPTIONS",
--- a/main.go
+++ b/main.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/ffuf/ffuf/pkg/encoder"
 	"github.com/ffuf/ffuf/pkg/ffuf"
 	"github.com/ffuf/ffuf/pkg/filter"
 	"github.com/ffuf/ffuf/pkg/input"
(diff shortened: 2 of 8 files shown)


bjhulst:
Running the encoders as part of ffuf will compromise a bit the performance... imho.

kazkansouh:
@bjhulst - Sure, this is true if you are dealing with a computationally bounded problem, such as running ffuf on an embedded device :confused:  

However, from what I have observed is that the limitation is more IO/network latency related, so even if a computationally demanding encoder was inline (e.g. perhaps something like building and signing a JWT) all that is needed is to increase the number of threads and the original throughput will be obtained albeit with more CPU usage.

Moreover, from my experience I often find myself using url encoding when fuzzing passwords with standard word lists and testing WAFs and have not noticed a performance hit, so I would say the usability improvement far outweighs this concern.

joohoi:
This will be mitigated in the near future by inclusion of https://github.com/ffuf/pencode functionality. 

I like the proposal for CLI syntax, and it will probably be adopted. pencode allows a lot of different encodings and chaining of them, so you will be able to do (for examples sake):
```
ffuf -w "wordlist.txt:FUZZ:urlencode b64encode" ...
```

resulting into urlencoded payload that's base64 encoded afterwards.

kazkansouh:
> This will be mitigated in the near future by inclusion of https://github.com/ffuf/pencode functionality.

@joohoi Thanks for the update. It will be nice to have this functionality!

joohoi:
Yeah, indeed. I feel kind of sorry because you put forward so much effort for this PR though :/

kazkansouh:
No problem, it was the best part of a day if I remember correctly. Feel welcome to reuse some of the code/take ideas this PR :wink: 

Should I close this PR?

Issue #227, opened 2020-04-27
Getting a single issue everytime i am running ffuf


Droidzzzio:
Just firing the normal basic command 
ffuf -u https://example.com/FUZZ -w wordlist.txt 

and getting an error

keyword defined, but not found in headers, method, URL or POST data

Using windows 10 
Problem occured after the updation to latest version of ffuf. 
Please resolve this issue ASAP

joohoi:
I'm unable to reproduce the issue, ffuf seems to work fine on windows 10 and with the latest release:
```
C:\Users\redacted\Downloads\ffuf_1.0.2_windows_amd64>type wordlist.txt
index.html
somethingelse
secret
C:\Users\redacted\Downloads\ffuf_1.0.2_windows_amd64>ffuf -u https://ffuf.io.fi/FUZZ -w wordlist.txt

        /'___\  /'___\           /'___\
       /\ \__/ /\ \__/  __  __  /\ \__/
       \ \ ,__\\ \ ,__\/\ \/\ \ \ \ ,__\
        \ \ \_/ \ \ \_/\ \ \_\ \ \ \ \_/
         \ \_\   \ \_\  \ \____/  \ \_\
          \/_/    \/_/   \/___/    \/_/

       v1.0.2
________________________________________________

 :: Method           : GET
 :: URL              : https://ffuf.io.fi/FUZZ
 :: Follow redirects : false
 :: Calibration      : false
 :: Timeout          : 10
 :: Threads          : 40
 :: Matcher          : Response status: 200,204,301,302,307,401,403
________________________________________________

secret                  [Status: 401, Size: 195, Words: 6, Lines: 8]:: Errors: 0 ::
index.html              [Status: 200, Size: 16, Words: 1, Lines: 4] :: Errors: 0 ::
:: Progress: [3/3] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
```

Common occurance is that people copy paste commands from a website where the dash `-` might be some other character instead, making ffuf not recognize the `-u` flag. I wonder if this might be t he issue?

Droidzzzio:
The issue is coming when i am firing ffuf command from the directory where the wordlist exist , . If the wordlist is in the other directory and giving upon the path , the error is ocurring .

Droidzzzio:
[...]

PR #240, opened 2020-05-16
Fixed behavior of wordlist:keyword separator in Windows

joohoi:
Fixes: #227


Twavesx:
in linux also there may be this exact same problem 
like the one i am facing with directories or files that contain ":" in their names

PR #262, opened 2020-07-13
Update random seed logic

bjhulst:
In GO rand.Intn is not really random at all unless you change the seed value before calling it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
     - New CLI flag `-maxtime-job` to set max. execution time per job.
     - Changed behaviour of `-maxtime`, can now be used for entire process.
   - Changed
-
+    - Fixed a bug related to the autocalibration feature making the random seed initialization also to take place before autocalibration needs it.
+    
 - v1.0.2
   - Changed
     - Write POST request data properly to file when ran with `-od`.
--- a/pkg/ffuf/job.go
+++ b/pkg/ffuf/job.go
@@ -331,6 +331,7 @@ func (j *Job) handleRecursionJob(resp Response) {
 //CalibrateResponses returns slice of Responses for randomly generated filter autocalibration requests
 func (j *Job) CalibrateResponses() ([]Response, error) {
 	cInputs := make([]string, 0)
+	rand.Seed(time.Now().UnixNano())
 	if len(j.Config.AutoCalibrationStrings) < 1 {
 		cInputs = append(cInputs, "admin"+RandomString(16)+"/")
 		cInputs = append(cInputs, ".htaccess"+RandomString(16))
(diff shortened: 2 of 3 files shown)


joohoi:
Thanks for the change. The `rand.seed` is however already called in `job.go`, as it only needs to be seeded once per execution (or job like here). It's much faster that way instead of applying a new seed for every call of the helper function.

This PR could however add a comment pointing that out for all the future readers of the code (myself included). For reference, the seed is set here: https://github.com/ffuf/ffuf/blob/7ffd74d87e15bf7673e323a02da5c4f7582580db/pkg/ffuf/job.go#L93

bjhulst:
Actually I found this issue when running with "calibration-enable" during "several times" and it always gave me the error message with the same random string.... :-)
 
Current code-base gives:
$ ./ffuf -ac -w words.txt -u https://notexisting.com/FUZZ
Error in autocalibration, exiting: Get https://notexisting.com/adminXVlBzgbaiCMRAjWw/: connect: no route to host
$ ./ffuf -ac -w words.txt -u https://notexisting.com/FUZZ
Error in autocalibration, exiting: Get https://notexisting.com/adminXVlBzgbaiCMRAjWw/:  connect: no route to host
$ ./ffuf -ac -w words.txt -u https://notexisting.com/FUZZ
Error in autocalibration, exiting: Get https://notexisting.com/adminXVlBzgbaiCMRAjWw/:  connect: no route to host

Adding the seed as in the PR.
$ vi pkg/ffuf/util.go 
$ go build
$ ./ffuf -ac -w words.txt -u https://notexisting.com/FUZZ
Error in autocalibration, exiting: Get https://notexisting.com/adminPFpLgfKsynySawSo/:  connect: no route to host

joohoi:
Good catch! You are right. 

The autocalibration happens before `job.Start()`. Perhaps a good place to do the random seed initialization would be https://github.com/ffuf/ffuf/blob/7ffd74d87e15bf7673e323a02da5c4f7582580db/main.go#L166 or just `main()`. I think the initialization in `job.Start()` should still stay, as the job runner can be used from external packages as well.

bjhulst:
I did some profiling on the amount of seed updates done when using this code from the PR. The code only calls the "seed update inside the RandomString function" 4 times per execution of the ffuf command. 

I think this current PR code will not harm performance and is cleaner to understand when the rand.Seed + rand.Intn are inside the same function block. :)

joohoi:
I agree that with the current functionality it's not big of an issue. However I think that seeding the random every time a random value is needed is kind of an antipattern, and would have to be changed around whenever the function gets more use. Because of this I think it'd be better to try and find the proper place to do the random seeding.

bjhulst:
How about updating the rand.Seed after https://github.com/ffuf/ffuf/blob/master/pkg/ffuf/job.go#L333  ?

bjhulst:
Is this good to go?

bjhulst:
CHANGELOG.md updated.

Issue #166, opened 2020-02-16
Disable bunch of errors in CLI


dhra:
Hi,
How can I disable errors in CLI results:
```bash
:: Progress: [646/6245] :: Job [1/1] :: 80 req/sec :: Duration: [0:00:08] :: Err:: Progress: [675/6245] :: Job [1/1] :: 84 req/sec :: Duration: [0:00:08] :: Err:: Progress: [687/6245] :: Job [1/1] :: 85 req/sec :: Duration: [0:00:08] :: Err:: Progress: [696/6245] :: Job [1/1] :: 87 req/sec :: Duration: [0:00:08] :: Err:: Progress: [708/6245] :: Job [1/1] :: 88 req/sec :: Duration: [0:00:08] :: Err:: Progress: [717/6245] :: Job [1/1] :: 89 req/sec :: Duration: [0:00:08] :: Err:: Progress: [727/6245] :: Job [1/1] :: 90 req/sec :: Duration: [0:00:08] :: Err:: Progress: [736/6245] :: Job [1/1] :: 81 req/sec :: Duration: [0:00:09] :: Err:: Progress: [745/6245] :: Job [1/1] :: 82 req/sec :: Duration: [0:00:09] :: Err:: Progress: [755/6245] :: Job [1/1] :: 83 req/sec :: Duration: [0:00:09] :: Err:: Progress: [765/6245] :: Job [1/1] :: 85 req/sec :: Duration: [0:00:09] :: Err:: Progress: [774/6245] :: Job [1/1] :: 86 req/sec :: Duration: [0:00:09] :: Err:: Progress: [783/6245] :: Job [1/1] :: 87 req/sec :: Duration: [0:00:09] :: Err:: Progress: [793/6245] :: Job [1/1] :: 88 req/sec :: Duration: [0:00:09] :: Err
```

joohoi:
That happens when writing a line to stderr (like the progress bar is), but the terminal is too narrow. For immediate fix, you should just try to make your terminal window wider. 

As a long term fix, ffuf should do terminal width detection and act accordingly by displaying a minimalist version of the progress bar.

I'm going to use this issue to track the development of this feature.

null-hyp0thesis:
To silence the Progress bar, just output stderr to /dev/null
`ffuf -u http://10.10.10.10 -w wordlist.txt 2>/dev/null`

[...]

PR #278, opened 2020-08-13
Fix: ffuf now minifies progress text if terminal width less than value.

Sam-Lane:
- ffuf/config.go added new variable to track if progress should be minified.
- ffuf/const.go added const to store value for minimum columns ffuf can handled before minifying.
- ffuf/util.go added functions to detect terminal size and to return true if size less than const MINIMUMCOLS
- output/stdout.go Progess function now changes what string should be printed based on Config.ProgessMinified
- main.go states to check if size of terminal is small and then assign alues to the config for the jobs.


Sam-Lane:
This fix is in regards to #166 

This is a work in progress as I need some feedback.
At the moment a const value stores the minimum columns before the output should be changed, currently set to 35. Should this be a smaller valuer? And should the minified progress output be even smaller?

joohoi:
Thanks for the (WIP) PR! This is a really nice improvement!

Few comments:
 - If by any means possible, I think it'd be beneficial to check if we need to minimize progress bar dynamically (until it has to be minimized). This should be done in the function that updates the progress bar. The reasoning behind this is that there are many variables affecting the current size - wordlist size (in # of digits), current position in the wordlist, number of jobs (in recursion), number of errors and such.
 - The terminal width check should be cross platform (I don't think Windows has `stty`), or the feature should be disabled for unsupported OSes (Windows)
 - I think the MVP (minimum viable progressbar, pun intended) would be job progress as percentage, and number of queued jobs. This is up for discussion however.

Sam-Lane:
I can make these changes. I imagine it should be if the string to be printed is at the length of the terminal only then print the minified version. 

It would be nice to constantly call stty so if say a user is running ffuf in tmux and rescaled the pane ffuf would update to compensate. But I considered about performance hit on that. I will implement when a user first runs ffuf it calculates the width and if the progress string will break that width it then minifies output. 

Regarding windows I'll ensure this feature is switched off for compatibility. 

Thanks for your kind feedback for this MVP for MVP :^)

Sam-Lane:
Here is an updated that covers your feedback.

progress is now dynamic switch back and forth between the minified version only if the verbose output does not fit the width of the terminal.

This feature is only enabled for non-windows machines. (tested on MacOS) If the host is a windows machine the width of the terminal is switched to 80 (the defacto).

The MVP Minimum Viable Progressbar now prints in the format of `[100%]-[1/1]` job progress as percentage and current job against total jobs. If this shorter version is still too long for a users terminal ffuf will then switch the `Config.Quite` flag to true which then turns off the printing of progress.

Line 175 of stdout.go
```go
progressString = fmt.Sprintf("%s[%.f%%%%]-[%d/%d]", TERMINAL_CLEAR_LINE, ffuf.Percent(status.ReqCount, status.ReqTotal), status.QueuePos, status.QueueTotal)
```
I had to double escape the `%` I'm not sure if this is intended or a bug on my part. It would be good if you could test this.

Let me know if any further changes need to happen.

Sam-Lane:
Let me know if there's any changes to be made @joohoi 👍🏻

joohoi:
I'm expecting to get this reviewed later today!

joohoi:
Oh, actually, meanwhile you could add yourself to CONTRIBUTORS.md and make a small addition to CHANGELOG.md

pkg/output/stdout.go
```diff
@@ -167,7 +168,20 @@ func (s *Stdoutput) Progress(status ffuf.Progress) {
...
-	fmt.Fprintf(os.Stderr, "%s:: Progress: [%d/%d] :: Job [%d/%d] :: %d req/sec :: Duration: [%d:%02d:%02d] :: Errors: %d ::", TERMINAL_CLEAR_LINE, status.ReqCount, status.ReqTotal, status.QueuePos, status.QueueTotal, reqRate, hours, mins, secs, status.ErrorCount)
+	//Create full progress string
+	progressString = fmt.Sprintf("%s:: Progress: [%d/%d] :: Job [%d/%d] :: %d req/sec :: Duration: [%d:%02d:%02d] :: Errors: %d ::", TERMINAL_CLEAR_LINE, status.ReqCount, status.ReqTotal, status.QueuePos, status.QueueTotal, reqRate, hours, mins, secs, status.ErrorCount)
+	if len(progressString)+1 >= s.config.TerminalWidth {
+		//if that progress string plus one is larger than the terminal width then replace the progress string with a shorter one
+		progressString = fmt.Sprintf("%s[%.f%%%%]-[%d/%d]", TERMINAL_CLEAR_LINE, ffuf.Percent(status.ReqCount, status.ReqTotal), status.QueuePos, status.QueueTotal)
+
+		if len(progressString) > s.config.TerminalWidth {
+			//finally if the new minified version of progress bar is too large to fit on the users terminal screen silence the progress
+			s.config.Quiet = true
```
joohoi, line 179: I think we should not set the Quiet here, but just return without printing the progress. The quiet mode alters the way the results are printed out as well, and this might not be favorable.

joohoi (changes requested):
The PR looks good! Only one minor thing that I'd like to see fixed.

However, I thought of the extensiveness of the information, and the end state I would like to have is having the current minimized version in this PR as the absolute "last resort". The work of adding different derivatives can be done either in this PR or a subsequent one.

The thing that I had in mind:

- Add information from "left to right"
- If even the current wordlist position does not fit, show the absolutely minimal version

In pseudocode:
```
if it fits, add wordlist position
  if it fits, add job count
     if it fits, add req/sec
       if it fits, add duration
         if it fits, add error counter 
else 
  if it fits, show minified status bar (as currently in this PR)
  else show nothing
```

What do you think?

joohoi:
Hi @Sam-Lane , sorry that it took so long for me to do the initial review and hence losing the momentum here. I think this is a brilliant addition, and would like to know if you'd like to continue working on it or if you wish to hand it over for me or someone else to continue the work on?

Sam-Lane:
Hei @joohoi 
Sorry I dropped ball on this. I just moved from finn to UK and started a new job. I'll be happy to finish this up soon. Once I'm settled 👌🏻

Do you have a release date in mind for this fix?

iDuronto:
This is an interesting addition. Any timeline when this will be merged? \o/

Sam-Lane:
Hi @joohoi,
Please find an update I believe fits your spec better.

joohoi:
Thanks for updating the PR. It looks good to me and works really well.

There seems to be a few merge conflicts right now, but the PR is ready to merge when those have been resolved.

joohoi (changes requested):
Thanks for updating the PR. It looks good to me and works really well.

There seems to be a few merge conflicts right now, but the PR is ready to merge when those have been resolved.

hakkuri01:
Still hitting this in v2.1.0-dev across multiple environments (Parrot OS, Fedora Linux boxes etc.). The fix in this PR looks complete per the last maintainer review. Is there anything blocking merge, or would it help if someone rebased this onto current master and resubmitted?

PR #289, opened 2020-09-03
Update default threads

shelld3v:
Brute forcing with over 40 threads will DoS the webserver, and will also make a very high chance to be detected and blocked by the server. Special when you use with the `-recursion` flag


codingo:
This pull request seems like a bit of a shot in the dark - what's the justification for 30 over 40?

joohoi:
I agree with @codingo here. There's no really safe value here, as depending on the connection and protocol (HTTP/2 vs. HTTP), the req/sec can be sky high with only a single thread too. I'm going to close this PR now, but feel free to comment here and we'll reopen it if needed.

Issue #141, opened 2020-01-25
Input command usability improvement suggestion


kazkansouh:
Hello, thank you for the tool - it is one of my go to tools :)

I have recently been using the `-input-cmd` function. However, I found it a bit of a challenge to find the correct interface to use (besides use of `$FFUF_NUM`).

This includes:
1. New lines in the output cause an error:

      ```
      $ ffuf -input-cmd 'echo $FFUF_NUM' -input-num=1 -u 'https://postman-echo.com/get?param=FUZZ' -v -debug-log /dev/stdout 
      ...snip...
      :: Progress: [0/1] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
      2020/01/25 08:14:37 parse https://postman-echo.com/get?param=0
      : net/url: invalid control character in URL
      :: Progress: [1/1] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 1 ::
      ```
    Compared to:

      ```
      $ ffuf -input-cmd 'echo -n $FFUF_NUM' -input-num=1 -u 'https://postman-echo.com/get?param=FUZZ' -v -debug-log /dev/stdout 
      ...snip...
      [Status: 200, Size: 226, Words: 5, Lines: 1]
      | URL | https://postman-echo.com/get?param=0
          * FUZZ: 1

      :: Progress: [1/1] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
     ```

    I think this is an acceptable issue (but could as well be easily fixed by just reading the first line), either way it would be helpful to have it mentioned it in the readme to save confusion.

2. The second issue is how the exit code is handled. I have noticed that in cases where the exit code is non-zero the output is silently ignored and the request is made. That is, consider the first example from above again but setting a non-zero exit code:

      ```
      $ ffuf -input-cmd 'echo $FFUF_NUM ; false' -input-num=1 -u 'https://postman-echo.com/get?param=FUZZ' -v -debug-log /dev/stdout 
      ...snip...
      [Status: 200, Size: 224, Words: 5, Lines: 1]
      | URL | https://postman-echo.com/get?param=
          * FUZZ: 1

      :: Progress: [1/1] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 0 ::
      ```

[...]

PR #293, opened 2020-09-12
Allow input shell option

jsgv:
* Add flag to set the shell to be used for the the input command

This is in response to issue #141 
Adds support to set the shell to be used with `input-cmd`. Fixes some issues that happen when using the [default shell](https://github.com/ffuf/ffuf/blob/master/pkg/input/const.go#L6) (`/bin/sh`). But still defaults to current shell if not set.

For example:
`echo` does not support `-n` flag in `/bin/sh`

Being able to set `/bin/zsh` or `/bin/bash` would fix the issue.

Usage:
```
ffuf -v \
    -input-shell "/bin/zsh" \
    -input-cmd 'echo -n subscriptions' \
    -input-num 1 -u https://example.com/path1/FUZZ
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -88,6 +88,7 @@ func main() {
 	flag.BoolVar(&conf.Colors, "c", false, "Colorize output.")
 	flag.BoolVar(&ignored, "compressed", true, "Dummy flag for copy as curl functionality (ignored)")
 	flag.Var(&opts.inputcommands, "input-cmd", "Command producing the input. --input-num is required when using this input method. Overrides -w.")
+	flag.StringVar(&conf.InputShell, "input-shell", "", "Shell to be used for running command")
 	flag.IntVar(&conf.InputNum, "input-num", 100, "Number of inputs to test. Used in conjunction with --input-cmd.")
 	flag.StringVar(&conf.InputMode, "mode", "clusterbomb", "Multi-wordlist operation mode. Available modes: clusterbomb, pitchfork")
 	flag.BoolVar(&ignored, "i", true, "Dummy flag for copy as curl functionality (ignored)")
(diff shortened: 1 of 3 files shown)


codingo (approved):

joohoi:
I do see the rationale behind this improvement, but I'm quite hesitant in adding new CLI flags to the already-too-long-to-read-and-understand usage listing.

My personal preference usage wise would be to `--input-cmd /path/to/script.sh` that has a shebang of `#!/path/to/preferred/shell` so the user could control it that way.

What do you think of this approach @jsgv ? I'm not saying no, I just want to understand the use case better to reach the best possible solution.

jsgv:
I believe the `input-cmd` is mostly used as a 1 liner. Anything more would probably make more sense to just generate a txt file and use as normal. 
I think that utilizing a shell file as you mentioned above is adding extra steps where it may make more sense to do it a different way.

Another approach is detecting the shell the current user is running and to default to that. What do you think about that?

joohoi:
I'm kind of conflicted here. I think that trying to automatically detect the active shell could cause more problems than it solves as it introduces unnecessary complexity. I do agree with your use case though. I think I'm ok with adding the new flag as initially proposed. 

The master branch went through some radical changes in between of getting this PR merged, so that'll need some work (sorry about that). I'd like you to add the new option to the example configuration file `ffufrc.example` as well.

joohoi:
Hi @jsgv , I wonder if you saw my last comment here :)

jsgv:
Yes, I saw it. I tried to fix the merge conflicts, but it was a bit overwhelming with all the updates and I did not have time to focus on it.

I think I will probably create a new PR from the latest master branch later.

jsgv:
Created new [PR](https://github.com/ffuf/ffuf/pull/344) 

This one can be deleted now if desired.

PR #308, opened 2020-09-27
Add support for configuration files for ffuf

joohoi:
This PR adds the possibility to set some of your commonly used configuration options to a configuration file. It also adds support for a default config file that gets applied for every ffuf job.


joakimtauren:
u mad bro

PR #312, opened 2020-09-28
Bug 231

bjhulst:
Hi, Would this solve bug 231?


joohoi:
This would prevent `-ac` of having any effect at all in case there already exists a filter of a kind. I think we should figure a way to append the `-ac` data to the existing filter instead.

bjhulst:
Adding the update part is a bit more work than this simple insert if statement. I will add it to this PR...

bjhulst:
Better now?

joohoi (changes requested):
That solution is really good tbh!

Just a small nitpick: the line indentations seem to be a bit off. You can automatically format the file using go: `go fmt pkg/filter/filter.go`. 

A tip: I have set my development environment to do this automatically on save, as a result I save a lot during development for auto formatting as I'm lazy ;)

If you also could add a short description of the fix to `CHANGELOG.md` it'd be great!

joohoi (approved):
LGTM!

PR #320, opened 2020-10-01
pkg: use {strings,bytes}.ReplaceAll when possible

### LINKED ISSUE #301: proposal: consider requiring Go 1.12 at least to simplify internal logic
Hi,

I propose to bump the minimum Go required version to, at least, Go 1.12 in order to be able to simplify some parts of the internal logic of `ffuf`.

### Context

`ffuf` [currently requires](https://github.com/ffuf/ffuf/blob/f3bcb50e3ad0960d810bbec0abe159bf0471704b/go.mod#L3) Go 1.11 at least. Go 1.12 was [released](https://blog.golang.org/go1.12) on 25 February 2019. As you can see on its [release notes](https://golang.org/doc/go1.12) two useful new functions were added to the standard libary. These are [`strings.ReplaceAll`](https://pkg.go.dev/strings#ReplaceAll) and [`bytes.ReplaceAll`](https://pkg.go.dev/bytes#ReplaceAll).

As you can see on their [implementation](https://golang.org/src/strings/strings.go?s=24475:24517#L958), this wouldn't have any performance benefit/consequence, it would just be a readability improvement.

### What would it simplify?

After doing a quick grep on ffuf's codebase, here's what could be simplified (commit [`f3bcb50e3ad0960d810bbec0abe159bf0471704b`](https://github.com/ffuf/ffuf/tree/f3bcb50e3ad0960d810bbec0abe159bf0471704b)):

```console
./pkg/runner/simple.go:75:		req.Method = strings.Replace(req.Method, keyword, string(inputitem), -1)
./pkg/runner/simple.go:78:			var CanonicalHeader string = textproto.CanonicalMIMEHeaderKey(strings.Replace(h, keyword, string(inputitem), -1))
./pkg/runner/simple.go:79:			headers[CanonicalHeader] = strings.Replace(v, keyword, string(inputitem), -1)
./pkg/runner/simple.go:82:		req.Url = strings.Replace(req.Url, keyword, string(inputitem), -1)
./pkg/runner/simple.go:83:		req.Data = []byte(strings.Replace(string(req.Data), keyword, string(inputitem), -1))
./pkg/filter/regex.go:44:		pattern = strings.Replace(pattern, keyword, regexp.QuoteMeta(string(inputitem)), -1)
```

### Implementation

I'd be happy to provide a patch doing it for Hacktoberfest if you agree on bumping the minimum Go required version 🎉 

Please, feel free to close this issue if you disagree (:

Best,
Miguel

jimen0:
### What does this PR do?

Now that Go 1.13+ is required, this PR makes use of standard library helper methods in order to have a more readable code.

### Problem description

See #301 

Fixes #301

### DIFF
--- a/pkg/filter/regex.go
+++ b/pkg/filter/regex.go
@@ -41,7 +41,7 @@ func (f *RegexpFilter) Filter(response *ffuf.Response) (bool, error) {
 	matchdata = append(matchdata, response.Data...)
 	pattern := f.valueRaw
 	for keyword, inputitem := range response.Request.Input {
-		pattern = strings.Replace(pattern, keyword, regexp.QuoteMeta(string(inputitem)), -1)
+		pattern = strings.ReplaceAll(pattern, keyword, regexp.QuoteMeta(string(inputitem)))
 	}
 	matched, err := regexp.Match(pattern, matchdata)
 	if err != nil {
--- a/pkg/runner/simple.go
+++ b/pkg/runner/simple.go
@@ -77,15 +77,15 @@ func (r *SimpleRunner) Prepare(input map[string][]byte) (ffuf.Request, error) {
 	req.Data = []byte(r.config.Data)
 
 	for keyword, inputitem := range input {
-		req.Method = strings.Replace(req.Method, keyword, string(inputitem), -1)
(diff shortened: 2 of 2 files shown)


joohoi:
LGTM!

PR #321, opened 2020-10-01
pkg/output/stdout: use Error() method from error interface when possible

jimen0:
### What does this PR do?

It simplifies error to string conversions by using the standard library [`error`](https://pkg.go.dev/builtin#error) interface `Error()` method.

### Why do we need it?

In my opinion, it improves the readability of the code. Feel free to close this PR if you don't think so.

Signed-off-by: Miguel Ángel Jimeno <miguelangel4b@gmail.com>


joohoi:
LGTM!

PR #322, opened 2020-10-01
pkg: handle gosimple linter findings

jimen0:
### What does this PR do?

It attempts to handle [`gosimple`](https://github.com/dominikh/go-tools/tree/master/simple) linter findings.

This change is an attempt to handle gosimple linter finfings in order to make the code easier to follow. It includes the following changes:

- use [`strings.Contains`](https://pkg.go.dev/strings#Contains) instead of `strings.Index` `!= -1`. As you can see in its [implementation](https://github.com/golang/go/blob/go1.15.2/src/strings/strings.go#L61), there's no performance improvement/consequence.
- use [`time.Since`](https://github.com/golang/go/blob/go1.15.2/src/time/time.go#L866-L867) standard library helper.
- remove unneeded return statements at the end of methods
- preallocate maps when their capacity is known [1]
- avoid underscoring values when they can be omitted. See this [playground](https://play.golang.org/p/UkgTavZIov3).
- avoid `fmt.Sprintf()` calls when the only argument is already a string

[1] In regards of the maps preallocation, it isn't that important as long as the maps aren't too big. Go maps are allocated with a small capacity if no size is provided. Then, their capacity is doubled after a certain load factor is reached. Growing their capacity means that, often, number of buckets must be doubled and entries need to be redistributed across them.

More information on how they're implemented can be found at: https://dave.cheney.net/2018/05/29/how-the-go-runtime-implements-maps-efficiently-without-generics
More information on how the load factor is chosen can be found at: https://github.com/golang/go/blob/a3868028ac8470d1ab7782614707bb90925e7fe3/src/runtime/map.go#L19L36

### Why do we need it?

It's part of my efforts to make ffuf simpler. In my opinion all simplifications improve readability and make contributors experience better. Please, feel free to close this PR if you disagree.

Best,
Miguel

### DIFF
--- a/pkg/ffuf/job.go
+++ b/pkg/ffuf/job.go
@@ -123,10 +123,7 @@ func (j *Job) Start() {
 }
 
 func (j *Job) jobsInQueue() bool {
-	if j.queuepos < len(j.queuejobs) {
-		return true
-	}
-	return false
+	return j.queuepos < len(j.queuejobs)
 }
 
 func (j *Job) prepareQueueJob() {
@@ -192,14 +189,13 @@ func (j *Job) startExecution() {
 	}
 	wg.Wait()
 	j.updateProgress()
-	return
 }
 
 func (j *Job) interruptMonitor() {
 	sigChan := make(chan os.Signal, 2)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 	go func() {
-		for _ = range sigChan {
+		for range sigChan {
 			j.Error = "Caught keyboard interrupt (Ctrl-C)\n"
 			j.Stop()
 		}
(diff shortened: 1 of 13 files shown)


jimen0:
Force-pushed as I accidentally changed the logic of `*WordlistInput.Next()` and `*CommandInput.Next()`. This PR should now be ready for review. Sorry about that. 🎉

joohoi:
LGTM, it apparently caught a merge conflight as I merged you previous PRs though. I'll merge when that's resolved.

jimen0:
Hi @joohoi,

I resolved the conflict now. It was the following change that caused it:
https://github.com/ffuf/ffuf/pull/322/files#diff-09d2a75209f409b024c87e33f513d4a9L81

```diff
-		headers := make(map[string]string, 0)
+		headers := make(map[string]string, len(req.Headers))
```

Fixed it by preserving the change that preallocated the map. Sorry about that.

Wish you a calmed Hacktoberfest 👍 I will try to come up with more and better contributions to your awesome project.

Best,
Miguel

joohoi:
Thank you! LGTM!

Inspired by this PR, I'm about to add a linter & code quality chercker as GitHub action to the repo.

joohoi (approved):

PR #330, opened 2020-10-14
Empty filter

bp0lr:
# Description

Hello, 
first, let me thank you for this amazing tool!

this small PR adds support to avoid creating empty files when you didn't get any result but the flags -of and -o was set.
It's become really annoying when you are doing a long recon, running ffuf again many urls, using many dictionaries, and empty results files are created just with the default template.

I'm not sure about the flag used (-or).
If you have a better option please let know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     - Ffuf now reads a default configuration file `$HOME/.ffufrc` upon startup. Options set in this file
     are overwritten by the ones provided on CLI.
     - Change banner logging to stderr instead of stdout.
+    - New CLI flag `-or` to avoid creating result files if we didn't get any. 
 
   - Changed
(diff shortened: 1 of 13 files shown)


joohoi:
Thanks for the PR! 

I initially thought I'd probably opt for having this on as default (and removing the CLI option altogether), but then again it might break some automated tools that use ffuf in their pipeline.

On the other hand if it's a cli parameter, It's easy to add to the default configuration when you know you want the behavior.

What do you think?

If you think it'd be beneficial to continue with having it as a CLI flag, I'd like you to add an example entry for it to the default example configuration file in https://github.com/ffuf/ffuf/blob/master/ffufrc.example as well.

bp0lr:
Maybe some tools use the created file to define a finished job.
I think you have a valid point about that.

I will keep the flag and add the new option to ffufrc.example.

joohoi (approved):
LGTM!

PR #335, opened 2020-10-18
pkg/ffuf: fix panic in Windows when parsing wordlist flag

### LINKED ISSUE #333: panic: runtime error: slice bounds out of range [:-1]
Hi pretty new to github so tell me if im missing things. i installed ffuf on windows but everytime i try to run it with a url and wordlist file i get this error.

C:\Users\xxx>ffuf -u google.com/FUZZ/ -w test.com
panic: runtime error: slice bounds out of range [:-1]

goroutine 1 [running]:
main.prepareConfig(0xc0000cc000, 0xc000058d00, 0x4, 0x7)
        /home/joona/go/src/github.com/ffuf/ffuf/main.go:314 +0x26f2
main.main()
        /home/joona/go/src/github.com/ffuf/ffuf/main.go:143 +0x19fb

I googled it but cant come up with a solution
Thank you

jimen0:
## Description

This change addresses two panics that happened while parsing the provided wordlist flag in Windows systems.

- [`pkg/ffuf/util.go:40`](https://github.com/ffuf/ffuf/blob/602c1c79e8399ccae5f6f7d645ab4f8e47e77118/pkg/ffuf/util.go#L34-L41): panic happened when the provided path was invalid. Example: `.\wordlist.txt:` as the `os.Stat` call returned an error different than `os.ErrNotExist`.

- [`pkg/ffuf/optionsparser.go:179`](https://github.com/ffuf/ffuf/blob/602c1c79e8399ccae5f6f7d645ab4f8e47e77118/pkg/ffuf/optionsparser.go#L179): panic happened when the provided value did not existed and did not contain a colon character. Example: `.\asdf.txt` when the local file `.\asdf.txt` did not exist. This panic happened due to `strings.LastIndex` returning `-1` when the provided substring does not appear. Therefore, `v[:-1]` panicking.

Fixes #333

## Additonally

---

**Please, test this patch thoroughly using a Windows system before merging it**. I'd like to contribute unit tests for this patch too in order to avoid having a regression or incomplete fix. However, it would require a bigger refactor of the file rather than just 2 panic fixes.

**Note**: [`pkg/ffuf/optionsparser.go:181`](https://github.com/ffuf/ffuf/blob/602c1c79e8399ccae5f6f7d645ab4f8e47e77118/pkg/ffuf/optionsparser.go#L181) (L185 on my patch) does not panic as `strings.LastIndex` returning -1 will make the slice access to be `v[0:]`, preserving the complete `v`value. It would be nice to prevent this in a next patch, though.

### DIFF
--- a/pkg/ffuf/util.go
+++ b/pkg/ffuf/util.go
@@ -31,11 +31,13 @@ func UniqStringSlice(inslice []string) []string {
 	return ret
 }
 
-//FileExists checks if the filepath exists and is not a directory
+//FileExists checks if the filepath exists and is not a directory.
+//Returns false in case it's not possible to describe the named file.
 func FileExists(path string) bool {
 	md, err := os.Stat(path)
-	if os.IsNotExist(err) {
+	if err != nil {
 		return false
 	}
+
 	return !md.IsDir()
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     - Fixed the issue where the option -ac was overwriting existing filters. Now auto-calibration will add them where needed.
     - The `-w` flag now accepts comma delimited values in the form of `file1:W1,file2:W2`.
     - Links in the HTML report are now clickable
+    - Fixed panic during wordlist flag parsing in Windows systems.
 
 - v1.1.0
   - New
--- a/pkg/ffuf/optionsparser.go
+++ b/pkg/ffuf/optionsparser.go
@@ -176,7 +176,11 @@ func ConfigFromOptions(parseOpts *ConfigOptions, ctx context.Context, cancel con
 				// The wordlist was supplied without a keyword parameter
 				wl = []string{v}
 			} else {
-				filepart := v[:strings.LastIndex(v, ":")]
+				filepart := v
+				if strings.Contains(filepart, ":") {
+					filepart = v[:strings.LastIndex(filepart, ":")]
+				}
+
 				if FileExists(filepart) {
 					wl = []string{filepart, v[strings.LastIndex(v, ":")+1:]}
 				} else {


pkg/ffuf/util.go
```diff
@@ -31,11 +31,13 @@ func UniqStringSlice(inslice []string) []string {
 	return ret
 }
 
-//FileExists checks if the filepath exists and is not a directory
+//FileExists checks if the filepath exists and is not a directory.
+//Returns false in case it's not possible to describe the named file.
 func FileExists(path string) bool {
 	md, err := os.Stat(path)
-	if os.IsNotExist(err) {
+	if err != nil {
```
joohoi, line 38: In what kind of scenario is this needed?

joohoi (changes requested):
Thanks for the PR! This looks solid. 

I would just like to understand the one bit I left a comment here (it's not a change request, but I would just like to understand the scenario here).

pkg/ffuf/util.go
jimen0: Hi @joohoi,

I documented it in the commit message too.

> panic happened when the provided path was invalid. Example: ".\wordlist.txt:" as the os.Stat call returned an error different than os.ErrNotExist.

This behavior only happens in Windows systems. Here is a reproducer you can use to test it (Windows only):

```go
package main

import (
	"fmt"
	"os"
)

func fileExists(path string) bool {
	md, err := os.Stat(path)
	if os.IsNotExist(err) {
		return false
	}

	return !md.IsDir()
}

func main() {
	fmt.Println(fileExists("/etc/passwd"))
	fmt.Println(fileExists("/thiswillneverexist"))
	fmt.Println(fileExists("/thispanicsinwindows:"))

}
```

And here's the output of running above program in Windows systems:

image

As you can see, `md.IsDir()` panics because `os.Stat` returned nil for `md` and an error different than `os.ErrNotExist` for `err`.

Please, let me know if you have further questions.

jimen0:
Hi @joohoi, did you have a chance to check my previous comment?

joohoi (approved):
Yeah, thanks a lot! LGTM, merging!

PR #344, opened 2020-11-10
Allow input-shell option

jsgv:
**This is a new pull request that replaces original [pull request](https://github.com/ffuf/ffuf/pull/293)**

* Add flag to set the shell to be used for the the input command

This is in response to issue #141 
Adds support to set the shell to be used with `input-cmd`. Fixes some issues that happen when using the [default shell](https://github.com/ffuf/ffuf/blob/master/pkg/input/const.go#L6) (`/bin/sh`). But still defaults to current shell if not set.

For example:
`echo` does not support `-n` flag in `/bin/sh`

Being able to set `/bin/zsh` or `/bin/bash` would fix the issue.

Usage:
```
ffuf -v \
    -input-shell "/bin/zsh" \
    -input-cmd 'echo -n subscriptions' \
    -input-num 1 -u https://example.com/path1/FUZZ
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     - Ffuf now reads a default configuration file `$HOME/.ffufrc` upon startup. Options set in this file
     are overwritten by the ones provided on CLI.
     - Change banner logging to stderr instead of stdout.
+    - New CLI flag `-input-shell` to set the shell to be used by `input-cmd`
 
   - Changed
     - Pre-flight errors are now displayed also after the usage text to prevent the need to scroll through backlog.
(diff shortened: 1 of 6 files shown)


joohoi (changes requested):
The code looks good. If you could add a short description to `CHANGELOG.md` and yourself to `CONTRIBUTORS.md` (alphabetically sorted), it'd be great!

jsgv (commented):
Added updates to markdown files

jsgv:
Anything else? @joohoi

joohoi (approved):

joohoi:
LGTM!

PR #373, opened 2021-01-10
Added 405 Method Not Allowed to the list of status codes matched by default

choket:
# Description

Added 405 Method Not Allowed to the list of status codes matched by default. A lot of APIs return 405 when accessing an endpoint using an invalid request method (GET instead of POST). Currently ffuf doesn't recognize that they are valid endpoints unless specifically told to match code 405.


joohoi:
LGTM!

joohoi (approved):

PR #397, opened 2021-02-18
Trim slash prefix

3lpsy:
# Description

This PR adds the option to automatically trim prefixes preceeding slashes from a wordlist. For example, if a wordlist contains:

```
/x.txt
profile
/admin
```
ffuf would make the following requests for `ffuf -u http://acme.com/FUZZ -w lists.txt -sp`:

```
/x.txt => http://acme.com/x.txt
profile => http://acme.com/profile
/admin => http://acme.com/admin
```

## Rationale

Typically, I would say that the wordlist should be prepared for the tool and context where the wordlist will be used. However, I have found myself creating countless duplicates of wordlists with the appendix "somelist-noslash.txt". I thought this would be a nice quality of life addition and save people time when switching to fresh VMs without having to load/prep custom wordlists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 - master
   - New
   - Changed
-  
+
+- v1.2.2
+  - New
+    - New CLI Flag `-sp` to automatically trim prefixed slashes in a wordlist
+ 
 - v1.2.1
   - Changed
     - Fixed a build breaking bug in `input-shell` parameter
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,4 +1,5 @@
 # Contributors
+* [3lpsy](https://github.com/3lpsy)
 * [AverageSecurityGuy](https://github.com/averagesecurityguy)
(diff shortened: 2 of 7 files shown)


joohoi:
Thanks for creating the PR.

I'm considering how to approach this. In the past I have been trying to avoid all features that can be handled "outside" of ffuf. I think this is important in order to avoid feature creep, and to be honest the amount of CLI flags available is making it harder to get a grasp of the tool and using it efficiently.

In my opinion, it would be a better fit for [ffuf-scripts](https://github.com/ffuf/ffuf-scripts) as a shell script (even though there's not much at the repo atm)

What do you think?

3lpsy:
I completely agree in keeping tools streamlined and this use case definitely does not fall under ffuf's domain so no worries there. It was just an experience I kept having moving from box to box. If it's an experience that many people are having, it might be worth making an exception, but for now, I'm okay with exploring external alternatives / leveraging bash i/o. Let me know what you think, but I'm fine letting the current PR fade into obscurity if there's no support for it.

PR #331, opened 2020-10-15
Dockerize all the things!

0x5ECF4ULT:
# Description

Hey. This is not a fix, instead I added a Dockerfile for your tool. Hope you like it. Now it's your turn to push it to Docker Hub ;-)

## Additonally

- [x] If this is the first time you are contributing to ffuf, add your name to `CONTRIBUTORS.md`. 
The file should be alphabetically ordered.
- [x] Add a short description of the fix to `CHANGELOG.md`

PR #402, opened 2021-02-27
add docker support

FireFart:
Hi,
this PR adds support for docker by adding a Dockerfile. You can create a project on docker hub and create autobuilds for this project now so the docker image will be up to date with every push to master.

I also added a basic makefile with some common commands

PS: the whitespace changes are from running `go fmt ./...`

Example run:
```bash
make docker-build
docker run --rm -it -v $(pwd)/Seclists:/wordlists ffuf:dev ffuf -u xxx -w /wordlists/Discovery/Web-Content/raft-large-directories.txt
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git
+_img
+.github
+*.md
+*.example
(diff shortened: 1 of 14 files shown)


bsysop:
Hi @FireFart,

There is a PR in the way, https://github.com/ffuf/ffuf/pull/331

Have you checked it? Probably you can improve there instead of a new PR.

PR #404, opened 2021-02-28
Fix Job - Recursion [Bug]

unKn0wnUser453:
Fix impossible condition in starting a recursion job
It now prints "[INFO] Scanning: http://URL/FUZZ" after the first job as intended.


joohoi:
Thanks for the PR! This was however fixed in the exclusive repository previously, so I'm going to have to close the pull request in order to avoid merge conflicts when bringing the changes back to the public repository in a week.

PR #405, opened 2021-02-28
Add ETC Counter [New Feature]

unKn0wnUser453:
This pull request adds an additional estimated time of completion (ETC) counter to the ffuf stdout functionality.

Let me know if this will suffice to be included to master, as I will probably be using ffuf with this feature from now on 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - master
   - New
      - All output file formats now include the `Content-Type`.
+     - Add an Estimated Time of Completion (ETA) to prompt.
   - Changed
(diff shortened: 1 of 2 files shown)


joohoi:
Thanks for the PR! 

I'll have to leave this to sit for a while, as there are currently issues with the status bar being too wide.

In general the feature is a good one, but there are a few things that I'd like to see taken into account:
- Include queued jobs in the estimation
- As the req/sec calculation for ffuf is close to realtime, and takes only the most recent requests into account and that value can fluctuate quite a bit, I fear that the time estimation jumps all around and might be hard to read for the user.

Issue #395, opened 2021-02-12
Create output dir recursively [essential when outputs parent folder does not exist)


Damian89:
Simple fix:

Instead of:
```
err := os.Mkdir(s.config.OutputDirectory, 0750)
```
use: 
```
err := os.MkdirAll(s.config.OutputDirectory, 0750)
```

Fix for:
```
-od /tmp/random/results
```

If "random" does not exist, the files wont get saved. Using MkdirAll will create "results" as well as "random".

Best,

Damian

joohoi:
Thanks for opening the issue. It should be an easy fix.

PR #413, opened 2021-03-16
Fix: Create output directory recursively instead of silently failing to produce output

mprencipe:
# Description

Output directory is now created recursively if the given directory path doesn't exist. Previously the output writing would silently fail in this case.

Fixes: #395 

## Additionally


joohoi (approved):
Thanks for the PR, LGTM!

Issue #259, opened 2020-06-27
Feature: Recursive brute force for a set of HTTP response codes


g33kyshivam:
Can we have a feature where a normal brute force is running based on HTTP response like 403/401 do a recursive brute force on that endpoint?

bsysop:
Hi @g33kyshivam, could you provide a example please?

g33kyshivam:
let's say that an endpoint /api returns a 401 response. So for my brute-force list if the response code is 401 then I want to launch a recursive brute force after /api/

so a command line switch would be great so specify these response codes which fuff would launch a recursive brute force say like

```fuff -responsecode 401,403```

the above should do a recursive search for that response status code

**Comment by bsysop** [comment]
If i'm not wrong, you want an additional config to preset to tell FFUF when should do some recursive work.

Today we have
```
  -recursion       Scan recursively. Only FUZZ keyword is supported, and URL (-u) has to end in it. (default: false)
  -recursion-depth Maximum recursion depth. (default: 0)
```

You wanna add something like

```
-recursion-if-code 401-403
```

Which would prevent FFUF of do recursion in `200` for example.

That is?

Today we have a simple check:
https://github.com/ffuf/ffuf/blob/7ffd74d87e15bf7673e323a02da5c4f7582580db/pkg/ffuf/job.go#L308-L310

`-mc/-fc` config is not what you want, but could help by now.

joohoi:
That's a good idea. The current recursion detection implementation is based on the ways the current httpd's handle directories. This proposal does make sense for API recursion though. 

I have previously thought that it's not that important, but I'm slowly changing my mind.

PR #416, opened 2021-03-28
add recursion-on-code flag

DeviousStoat:
# Description

I think I added the feature requested in #259.
It adds a flag `-recursion-on-code` expecting comma separated HTTP status codes or ranges. Then the recursion will only take place on directories returning status codes matching those `-recursion-on-code` codes.

Fixes: #259 

## Additonally

### DIFF
--- a/help.go
+++ b/help.go
@@ -54,7 +54,7 @@ func Usage() {
 		Description:   "Options controlling the HTTP request and its parts.",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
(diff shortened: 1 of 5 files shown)


joohoi:
There's a new CLI flag `-recursion-strategy` in the exclusive repository currently, that is going to be merged in the public repository in a week. This flag has a value `greedy` that adds a new recursion job on every hit. 

I like the idea in this PR and would suggest rebasing it on the new feature, possibly adding a new `recursion-strategy` value: `code` to it.

PR #424, opened 2021-04-08
Add flag to ignore certain words from wordlist

KwnyPwny:
# Description

This pull requests adds the flag `-iw`.
The flag accepts a list of comma separated values, like the extension flag `-e`.
When loading a wordlist, it is checked for every word if it contains a value from the `-iw` flag.
This can be shown most easily with an **example**:
File `words`:
```
dologout
logarithm
login
logout
logoutnow
logs
```

Cmd: `ffuf -u 'https://www.google.com:443/FUZZ?a=1&b=2' -w words -iw logout`

Words that are loaded:
```
logarithm
login
logs
```

## Why is this useful?

As the example above demonstrates it is, for example, possible to remove the word "logout" from wordlists.
This has the advantage, that when doing authenticated fuzzing, you are not logged out because the `/logout` endpoint is not called.
This is currently only possible by copying the wordlist and removing the unwanted words, which I find inconvenient.

I'm happy about feedback :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - master
   - New
      - All output file formats now include the `Content-Type`.
+     - New CLI flag `-iw` to provide a list of substrings, which, if present within a words from the wordlist (`-w`), exclude this word from the tests. 
   - Changed
     - Fix a badchar in progress output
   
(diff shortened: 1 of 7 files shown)


joohoi:
Thanks for creating the PR.

I'm considering how to approach this. In the past I have been trying to avoid all features that can be handled "outside" of ffuf. I think this is important in order to avoid feature creep, and to be honest the amount of CLI flags available is making it harder to get a grasp of the tool and using it efficiently.

In my opinion, it would be a better fit for [ffuf-scripts](https://github.com/ffuf/ffuf-scripts) as a shell script (even though there's not much at the repo atm)

What do you think?

KwnyPwny:
Hi, thanks for your feedback. I absolutely understand your mindset on this. Closing this PR.

Issue #440, opened 2021-04-30
Add TLS SNI Support


erbbysam:
As an end user I would like to run ffuf with a TLS SNI I specify. This is important in 2 cases:
1) I do a DNS query in advance (avoiding a duplicate DNS query)
2) I'm attempting to run ffuf against a hostname/IP pair that DNS does not resolve

### Examples
This request will run with no TLS SNI (also see https://github.com/golang/go/issues/22704):
```
./ffuf -H "Host: www.erbbysam.com" -u https://104.154.120.133:443/FUZZ -w ~/wordlist.txt
```

This request will have a TLS SNI, but create an extra DNS lookup:
```
./ffuf -u https://www.erbbysam.com443/FUZZ -w ~/wordlist.txt
```

### Proposal
Add a new `-sni` flag that will be used when establishing TLS connections:
```
./ffuf -H "Host: www.erbbysam.com" -u https://104.154.120.133:443/FUZZ -sni www.erbbysam.com -w ~/wordlist.txt 
```

I will send out a PR shortly with support for this flag.

PR #441, opened 2021-04-30
Add TLS SNI Support

erbbysam:
# Description

Add TLS SNI support to ffuf.

I opted not to include support for the `FUZZ` target here as each new request would require a full new TLS connection which introduces significant overhead. There might be some value in adding support for this later for VHOST discovery using TLS SNI.

Fixes: #440

For background: https://tools.ietf.org/html/rfc6066#section-3

## Example:
```
./ffuf -H "Host: www.erbbysam.com" -u https://104.154.120.133:443/FUZZ -sni www.erbbysam.com -w ~/wordlist.txt 
```

## Tests:
 - I have confirmed this new field works.
 - I have confirmed this new code does not impact existing SNI values when run without the `-sni` flag using.
 - I have confirmed this new field can be specified via configuration file.

(all tests were done using using tcpdump+wireshark -- example: `tcpdump port 443 -w ~/output.pcap`)

## Additonally

- [X] If this is the first time you are contributing to ffuf, add your name to `CONTRIBUTORS.md`. 
- [X] Add a short description of the fix to `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 - master
   - New
+    - Added a CLI flag to specify TLS SNI value
   - Changed
 
 - v1.3.1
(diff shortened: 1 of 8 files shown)


joohoi (approved):
Thanks for the PR, LGTM!

PR #444, opened 2021-05-13
Fix an issue where output file was created regardless of -or

joohoi:


pkg/output/stdout.go
```diff
@@ -272,6 +268,10 @@ func (s *Stdoutput) writeToAll(filename string, config *ffuf.Config, res []ffuf.
 // SaveFile saves the current results to a file of a given type
 func (s *Stdoutput) SaveFile(filename, format string) error {
 	var err error
+	if s.config.OutputSkipEmptyFile && len(s.Results) == 0 {
```
hpy, line 271: The length of `s.Results` is not guaranteed to have the results stored in it at the time of this check. This is because they may still be sitting in `s.CurrentResults` having not been appended. This means when we use the `-or` argument it will always fail to create a file!

We need to check the length of the result from `append(s.Results, s.CurrentResults...)` as you were doing previously inside each `writeJSON`, `writeEJSON`... 

nit: it would be simpler to have one line like:
```
res = append(s.Results, s.CurrentResults...)
```
and use the `res` as an arg to pass to all the output functions.

PR #469, opened 2021-07-13
Add Sniper Mode

denandz:
# Description

This pull request adds a "sniper" mode to ffuf. This lets you specify payload locations using template characters, and ffuf will then run through each one. TLDR, this now works:

```
./ffuf -w wordlist -u http://127.0.0.1:8000/test?feature=§aaa§&thingie=§bbb§&array[§0§]=§baz§ --mode sniper
```

ffuf will hit each of the payload locations above, one after the other, while preserving the other values.

Here's some background before delving into the details.

The main issue is that using ffuf to hunt for vulnerabilities rather than do content discovery can be rather clunky. What I'd like to do is take a single request that I'm interested in, highlight the locations I want to attack, and then have ffuf go through and smash my word-list at each value, one after the other, while preserving the legitimate values in the other fields I've highlighted.

Currently, a single request would need multiple command line runs and multiple request files (if you use ffuf with request files instead of building out every request by hand with flags), one for each fuzzing location. For example, the following request (an attempt to log in to GitHub):

```
POST /session HTTP/1.1
Host: github.com
...snip...

commit=Sign+in&authenticity_token=...redact..&login=foo&password=bar&trusted_device=&webauthn-support=supported&webauthn-iuvpaa-support=unsupported&return_to=..redact..&allow_signup=&client_id=&integration=&required_field_9b00=&timestamp=..redact..&timestamp_secret=..redact..
```

This would need a 14 different request files for just value side of the `POST` parameters, each with a `FUZZ` parameter. When looking at an entire application this becomes problematic. With the sniper mode that this pull request implements, instead of using 14 different request files, there is just one:

```
POST /session HTTP/1.1
Host: github.com
...snip...

commit=§Sign+in§&authenticity_token=§...redact..§&login=§foo§&password=§bar§&trusted_device=§§&webauthn-support=§supported§&webauthn-iuvpaa-support=§unsupported§&return_to=§..redact..§&allow_signup=§§&client_id=§§&integration=§§&required_field_9b00=§§&timestamp=§..redact..§&timestamp_secret=§..redact..§
```

The section sign characters can be added in `vi` using `ctrl-k` followed by `shift+S+E`. There are key combos for windows/macos/linux/whatever detailed [here](https://en.wikipedia.org/wiki/Section_sign)

This pull request touches a fair bit of ffuf's internals, which I'll try go over in some detail.

## General arch

I tried to integrate these changes with the existing ffuf design without changing the internals too much wherever possible. The new sniper mode is a thin veil over the clusterbomb mode. It generally works like this:

- The user specifies sniper mode, and the options parser does some checks to make sure the arguments are sane.
- The input word-list is loaded as per the clusterbomb mode, with `FUZZ` used as the keyword
- The job runner logic starts (`Start()`) and if sniper mode is configured, then the requests are parsed for template characters (`§`)
- An array of requests is returned by `SniperRequests`, with one location specified by the user replaced by `FUZZ` in each request.
- A queue job is created per request
- The job runner continues as-per usual

## Change summary

- `main.go` - minor changes for new mode string
- `input/input.go` - minor changes to support the new mode string
- `interactive/termhandler.go` - Queue jobs are no longer just for recursion jobs, so the language has been changed in the termhandler to reflect that.
- `runner/simple.go` - `Prepare()` has been modified to take a Request parameter, rather than build a new `Request` struct from the global config.
- `ffuf/config.go` - `InputProviderConfig` now includes a `Template` string field which holds the payload locator symbol used by sniper mode.
- `ffuf/interfaces.go` - `RunnerProvider` has been tweaked such that `Prepare` now requires a base request.
- `ffuf/job.go` - `QueueJob` structs now hold a `Request` field, this is the base request used by each QueueJob when it's running. `Start()` has been modified to implement the sniper request parser when necessary, and additional changes made to the other methods to support pulling the base request for each job out of the queue struct.
- `ffuf/optionsparser.go` - Changes for the additional mode and a validator for sniper configuration. This now handles the modes prior to dealing with wordlists/input commands.
- `ffuf/optionsparser_test.go` - New file, implements test logic for the sniper validator.
- `ffuf/request.go` - New methods to generate requests from the current config, copy requests and handle sniper request parsing logic.
- `ffuf/request_test.go` - New file, tests for the new Request functions


joohoi (changes requested):
Thanks for the PR! Sorry that it took so long to get around to review this. This is a great addition, and a total beast of a PR 💥

I love how you added tests alongside of it too. 

I'm going to need a bit more time to go through it all. In case you haven't completely given up on this because of the long delays, maybe you could fix the merge conflict I must have caused with other changes meanwhile?

denandz:
Heya, definitely still keen to get this merged. I'll take a look at the merge conflicts during the week and see if i can sort it out.

denandz:
Looks like the merge conflict was just the language changes in the termhandler. Easy fix

pkg/ffuf/job.go
```diff
@@ -107,10 +108,22 @@ func (j *Job) Start() {
 		j.startTime = time.Now()
 	}
 
-	// Add the default job to job queue
-	j.queuejobs = append(j.queuejobs, QueueJob{Url: j.Config.Url, depth: 0})
+	basereq := BaseRequest(j.Config)
+
+	if j.Config.InputMode == "sniper" {
+		// process multiple payload locations and create a queue job for each location
+		reqs := SniperRequests(&basereq, "§")
```
joohoi, line 115: I assume the template string should be fetched from the `InputProvider` where we set it up in `optionsparser.go`, making it easier to change later on or maybe even added as user configurable value in the future.

pkg/ffuf/request_test.go
```diff
@@ -0,0 +1,246 @@
+package ffuf
```
joohoi, line 1: I love the added tests <3

joohoi (changes requested):
All in all, looks good! I absolutely love that you included tests for the couple heavily modified files too. The only minor tweaks that came to my mind were reading the template char from the inputprovider itself, as well as the following:

I wonder if it would make sense to reword this job-start message in case the user is running in sniper mode: 

https://github.com/ffuf/ffuf/blob/b56de007d4571de00491350348f35a8186cb4546/pkg/ffuf/job.go#L208

pkg/ffuf/job.go
denandz: Good catch! Should be an easy fix

pkg/ffuf/request_test.go
denandz: do they.... pass the test?

image

denandz:
Righto, I've updated sniper mode to use the `InputProvider` for the template character. Currently its pulling it from the first `InputProvider`, which given that the section sign character is currently hard-coded shouldn't be a big deal. 

Changing the job-start message is a good idea, I've made that change so the user can see how far through the sniper job queue they've progressed

denandz:
Heya, keen to get this merged. Was there any other info or changes you needed from me?

joohoi (approved):
Yeah, it's good to go, sorry for the delay!

harmtemolder:
I saw `sniper` in `ffuf -h` and read through this PR to understand what it does. This might come in handy!

@denandz Just a suggestion, but it might be useful to add it to https://github.com/ffuf/ffuf/wiki as well.

Issue #357, opened 2020-11-25
wordlists not readable in ffufrc 


0xspade:
Hi @joohoi,

I set up a large file of wordlists in .ffufrc file but when I'm using the command:
```bash
$ ffuf -u https://example.com/FUZZ
```
it says that I need to provide a wordlist which I already did in .ffufrc file.

joohoi:
Could you please include the content of your `~/.ffufrc` file for debugging purposes? Also just to make sure: the file is in the `$HOME/.ffufrc` right?

0xspade:
Hi @joohoi,

here is my `.ffufrc`:

```yaml
<snip...>
[input]
        dirsearchcompat = true
        extensions = "conf,config,bak,backup,swp,old"
        ignorewordlistcomments = true
        wordlists = [
                "/path/dicc.v3.txt:FUZZ",
                "/path/assetnote-wordlists/bak.txt:BAK"
       ]
<snip...>
```

KwnyPwny:
I'm having the same issue on v1.3.1.

paupaulaz:
Same here :(

Issue #484, opened 2021-09-15
Wordlist args missing when referenced from config file.


jamf25:
Didn't want to have to create github issue, but I don't see anywhere else discussing this problem.  When using the -config option the wordlists declared in the config appear to be ignored.  

**Config stanza (copied from template)**
```
[input]
    dirsearchcompat = false
    extensions = ".php,.txt,.bak,.swp,.js"
    ignorewordlistcomments = false
    inputmode = "clusterbomb"
    inputnum = 100 
    inputcommands = [ 
    #    "seq 1 100:CUSTOMKEYWORD"
    ]   
    #request = "requestfile.txt"
    requestproto = "http"
    wordlists = [ 
        "/seclists/Discovery/Web-Content/raft-medium-words.txt:FUZZ"
    ]
```

**command using referenced config file**
`ffuf -config ~/.xffufrc -u http://[url]/FUZZ`

**resulting output**
```
More information and examples: https://github.com/ffuf/ffuf

Encountered error(s): 1 errors occurred.
        * Either -w or --input-cmd flag is required
```

I've looked at other directives in the config file that may suggest something about ignoring wordlists or something but have not found anything.

jamf25:
Appears to be a dub of #357  apologies.

PR #488, opened 2021-09-23
Deleted Re-parse the cli options from main.go

xaoirse:
It set opts.Input.Wordlists to []. and caused such errors:
Wordlist args missing when referenced from config file. #484
wordlists not readable in ffufrc #357

Fixes: #357 #484


joohoi:
Thanks for your attempt to fix the issue. Re-parsing the commandline is however needed in order to override some values defined in the ffufrc file, so the commandline parameters take priority over the "placeholder" values defined in the ffufrc file. 

Closing this PR in favor of #511 that addresses this aspect as well.

PR #499, opened 2021-11-17
Update README.md

adilsoybali:
# Description

Installing executables with "go get" in module mode is deprecated.
"go install pkg@version" should be used instead.
For more information, see https://golang.org/doc/go-get-install-deprecation


CONTRIBUTORS.md
```diff
@@ -35,3 +35,4 @@
 * [SolomonSklash](https://github.com/SolomonSklash)
 * [l4yton](https://github.com/l4yton)
 * [xfgusta](https://github.com/xfgusta)
+* [adilsoybali](https://github.com/adilsoybali)
```
joohoi, line 38: This list should be kept in alphabetically sorted order, can you make that change?

joohoi (changes requested):
Thanks for the PR! This is a good documentation change. 

However, I think the required Go version mentioned [here](https://github.com/ffuf/ffuf/blob/master/README.md?plain=1#L50) should be changed to 1.16 at the same time in order to not to cause confusion if `go install` doesn't work as expected.

adilsoybali:
Hi @joohoi. I made the changes you requested. If there's anything else that needs to change, I can do that too. Thank you, have a nice day.

joohoi (approved):
Thanks for the quick response. It's all good now, merging!

PR #510, opened 2022-01-20
500 Internal Server Error -> default matched status codes

noraj:
# Description

Add 500 Internal Server Error to list of status codes matched by default

# Why?

500 status code is frequently observed when the app doesn't behave normally which is often interesting for a security auditor

Not having 500 matched by default may let us miss some important stuff

During an assessment I first miss a 500 status with teh default matcher I then found with `-mc all -fc 400`.

There is no option to **add** a matching status code (matching += 500) so either we have to do something like  `-mc all -fc 400` or to specify the full matching list again eg. `-mc 200,204,301,302,307,401,403,405,500`. That why having 500 in the default list, as 403 and 401 if it generates too much false positives it's easier to have it by default and exclude it with `-fc 500` then.

I added it it in my `~/.ffufrc` but I think it's better to have it by default.


joohoi (approved):
Thanks for the pull request, I agree with your reasoning here, and the PR looks good

PR #519, opened 2022-02-14
Update README.md

wellencamass:
make examples for readme

# Description

Please add a short description of pull request contents.
If this PR addresses an existing issue, please add the issue number below.

Fixes: #(issue number)

## Additonally


joohoi (changes requested):
The text in the changed spots is actually pulled from the ffuf help text. I also think that it would probably be more efficient to write about the different filtering options in a more elaborate way and maybe under a specific subtitle in the README instead.

Issue #377, opened 2021-01-18
recursion not working properly


cardassian-tailor:
Hello, even with -recursion enabled, and -recursion-depth of 3 ... and im using FUZZ at the end of both the URL and the wordlist ... I am able to locate directories but new jobs are not added for the recursion process. 

So - i have to re-run ffuf and manually scan those new directories. 


```
ffuf -w /opt/SecLists/Discovery/Web-Content/raft-medium-directories.txt:FUZZ -u http://masked.masked/FUZZ -r -recursion -recursion-depth 2 -o ffuf_scan -of all -e .php,.aspx,.txt,.md  
```

```
:: Method           : GET                                                                                                                                                                                                                                               
 :: URL              : http://masked.maskedFUZZ                                                                                                                                                                                                                           
 :: Wordlist         : FUZZ: /opt/SecLists/Discovery/Web-Content/raft-medium-directories.txt                                                                                                                                                                             
 :: Extensions       : .php .aspx .txt .md                                                                                                                                                                                                                               
 :: Output file      : ffuf_scan.{json,ejson,html,md,csv,ecsv}                                                                                                                                                                                                           
 :: File format      : all                                                                                                                                                                                                                                               
 :: Follow redirects : true                                                                                                                                                                                                                                              
 :: Calibration      : false                                                                                                                                                                                                                                             
 :: Timeout          : 10                                                                                                                                                                                                                                                
 :: Threads          : 40                                                                                                                                                                                                                                                
 :: Matcher          : Response status: 200,204,301,302,307,401,403                                                                                                                                                                                                      
________________________________________________                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                         
test                    [Status: 200, Size: 47066, Words: 2465, Lines: 651]                                                                                                                                                                                              
test.php                [Status: 200, Size: 47078, Words: 2465, Lines: 651]                                                                                                                                                                                              
index                   [Status: 200, Size: 177, Words: 22, Lines: 5]                                                                                                                                                                                                    
secret                 [Status: 200, Size: 11406, Words: 1103, Lines: 294]                                                                                                                                                                                              
                        [Status: 200, Size: 177, Words: 22, Lines: 5]                                                                                                                                                                                                    
rename                  [Status: 200, Size: 95, Words: 4, Lines: 1]                                                                                                                                                                                                      
[...]

PR #522, opened 2022-02-19
Fix a recursion bug (#377)

h1x-lnx:
Recursion does not work when the user provides port information in the URL and the site returns redirection status code (3xx) with a `Location` header containing absolute URL path. Bug (#377).

Line responsible: https://github.com/ffuf/ffuf/blob/965f282c0b77b620ed418a5905d10a171f627448/pkg/ffuf/response.go#L46

This line drops the port info when `redirectUrl` is an absolute path and is the same as `baseUrl`.
This PR fixes the bug by making sure that the above situation is handled differently than when the `redirectUrl` contains a relative path.


joohoi (approved):
Thanks for the PR, LGTM!

PR #570, opened 2022-07-12
Change precedence of quiet and JSON output to favour JSON

tomnomnom:
This is a pretty small change that enables something I think is useful.

At the moment if you want all three of:

* JSON output on `stdout`
* No banner / progress bar
* To still get errors on `stderr`

There's no easy way to do it. If you redirect `stderr` to `/dev/null` you lose the error messages, if you use `-s`/silent/quiet mode you lose the JSON output.

This change in precedence allows you to use both the `-s` and `-json` flags to achieve the above, and shouldn't break any existing use-cases as the `-json` flag currently does nothing if you're also using the `-s` flag.


joohoi:
Sorry for the long wait in the queue. LGTM and merging now.

tomnomnom:
Thank you for the merge! :)

Issue #573, opened 2022-07-27
ffuf silently ignores invalid proxy options


denandz:
Setting either the `-x` or `--replay-proxy` options to some incorrect value results in `ffuf` running and ignoring the proxy value. This lead to me accidentally putting `http:/127.0.0.1:8080` as the replay proxy and then spending the next 20 minutes questioning the fabric of reality.

Some validation of these flags and an error when they're invalid would be a good addition.

```
$ ~/go/bin/ffuf -x "asdfasdf asf" --replay-proxy "asdfasdf qwer"  -w test -u http://127.0.0.1:8000/FUZZ -mc all

        /'___\  /'___\           /'___\       
       /\ \__/ /\ \__/  __  __  /\ \__/       
       \ \ ,__\\ \ ,__\/\ \/\ \ \ \ ,__\      
        \ \ \_/ \ \ \_/\ \ \_\ \ \ \ \_/      
         \ \_\   \ \_\  \ \____/  \ \_\       
          \/_/    \/_/   \/___/    \/_/       

       v1.5.0-dev
________________________________________________

 :: Method           : GET
 :: URL              : http://127.0.0.1:8000/FUZZ
 :: Wordlist         : FUZZ: test
 :: Follow redirects : false
 :: Calibration      : false
 :: Proxy            : asdfasdf asf
 :: ReplayProxy      : asdfasdf qwer
 :: Timeout          : 10
 :: Threads          : 40
 :: Matcher          : Response status: all
________________________________________________

:: Progress: [4/4] :: Job [1/1] :: 0 req/sec :: Duration: [0:00:00] :: Errors: 4 ::
```

PR #574, opened 2022-07-29
Added additional proxy URL verification

denandz:
# Description

This pull request adds additional checking for proxy and replay-proxy parameters, as well as adding some tests to make sure the parser is working how it's expected to.  Go's `url.parse` alone does not provide enough verification to determine whether a given URL matches the requirements for a proxy passed to a HTTP transport.

Fixes: #573

### DIFF
--- a/pkg/ffuf/optionsparser.go
+++ b/pkg/ffuf/optionsparser.go
@@ -364,9 +364,9 @@ func ConfigFromOptions(parseOpts *ConfigOptions, ctx context.Context, cancel con
 
 	// Verify proxy url format
 	if len(parseOpts.HTTP.ProxyURL) > 0 {
-		_, err := url.Parse(parseOpts.HTTP.ProxyURL)
-		if err != nil {
-			errs.Add(fmt.Errorf("Bad proxy url (-x) format: %s", err))
+		u, err := url.Parse(parseOpts.HTTP.ProxyURL)
+		if err != nil || u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "socks5") {
+			errs.Add(fmt.Errorf("Bad proxy url (-x) format. Expected http, https or socks5 url"))
 		} else {
 			conf.ProxyURL = parseOpts.HTTP.ProxyURL
 		}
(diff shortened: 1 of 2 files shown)


pkg/ffuf/optionsparser.go
```diff
@@ -364,19 +364,19 @@ func ConfigFromOptions(parseOpts *ConfigOptions, ctx context.Context, cancel con
 
 	// Verify proxy url format
 	if len(parseOpts.HTTP.ProxyURL) > 0 {
-		_, err := url.Parse(parseOpts.HTTP.ProxyURL)
-		if err != nil {
-			errs.Add(fmt.Errorf("Bad proxy url (-x) format: %s", err))
+		u, err := url.Parse(parseOpts.HTTP.ProxyURL)
+		if err != nil || u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "socks5") {
```
joohoi, line 368:
```suggestion
		if err != nil || u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "socks5" && u.Scheme != "socks5h") {
```
Even though they're [handled similarly by the Go net/proxy](https://cs.opensource.google/go/x/net/+/refs/tags/v0.5.0:proxy/proxy.go;l=92-99), there is an additional valid flavor of `socks5` called `socks5h`, so in the spirit of completeness, I think we should allow that too.

pkg/ffuf/optionsparser.go
```diff
@@ -364,19 +364,19 @@ func ConfigFromOptions(parseOpts *ConfigOptions, ctx context.Context, cancel con
...
 		}
 	}
 
 	// Verify replayproxy url format
 	if len(parseOpts.HTTP.ReplayProxyURL) > 0 {
-		_, err := url.Parse(parseOpts.HTTP.ReplayProxyURL)
-		if err != nil {
-			errs.Add(fmt.Errorf("Bad replay-proxy url (-replay-proxy) format: %s", err))
+		u, err := url.Parse(parseOpts.HTTP.ReplayProxyURL)
+		if err != nil || u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "socks5") {
```
joohoi, line 378:
```suggestion
		if err != nil || u.Opaque != "" || (u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "socks5" && u.Scheme != "socks5h") {
```
Even though they're [handled similarly by the Go net/proxy](https://cs.opensource.google/go/x/net/+/refs/tags/v0.5.0:proxy/proxy.go;l=92-99), there is an additional valid flavor of `socks5` called `socks5h`, so in the spirit of completeness, I think we should allow that too.

joohoi (changes requested):
LGTM with this proposed small nit

denandz:
> LGTM with this proposed small nit

good catch! Committed.

joohoi (approved):
Thanks for the quick fix!

Issue #584, opened 2022-08-25
maxtime should save output when timeouts


dpanic:
I think maxtime should save output when process context timeouts.

For now I made workaround...
1. Started process externally, spawned it
2. Every 60 seconds, my thread writes to stdin of process
```
\n
savejson file.json\n
resume\n
```

This is not solution ...

Version: 1.5.0-dev

PR #585, opened 2022-08-29
chore: implement save output on maxtime reached

dpanic:
# Description

Minor change, saves output if maxtime is reached.

Fixes: #584 

## Additonally
* [x]  If this is the first time you are contributing to ffuf, add your name to `CONTRIBUTORS.md`.
* [x]  Add a short description of the fix to `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - master
   - New
   - Changed
+    - Fixes issue where maxtime stops process without saving output
   
 - v1.5.0
   - New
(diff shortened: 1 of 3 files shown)


dpanic:
@joohoi did you had time to look at this?

dpanic:
Any updates on this?

wellenc-lex (approved):
we need this!

joohoi:
Sorry for taking such a long time to get around this PR. I'm a bit confused about it when it comes to the core issue at hand. 

The output should already be saved whenever any of the exit conditions are met. This should happen because whenever `Stop()` is called within the loop that runs the job (and any recursion jobs), is just simply iterated through without actual execution, and the output file save happens directly after before actually quitting.

https://github.com/ffuf/ffuf/blob/master/pkg/ffuf/job.go#L139-L149

If you have encountered a situation where the output file has _not_ been generated, we need to dig into this further as it most likely is a bug somewhere else in the code than a missing attempt to save the file.

Issue #571, opened 2022-07-16
Feature: Stop on first successfull response


wellencamass:
If you bruteforce HTTP Basic you dont need to continue sending new requests if you have found valid credentials pair so it would be cool to add feature that stops ffuf after first successfull ( with -fc 403 ) response

opabravo:
Bump

PR #586, opened 2022-09-07
Add -ss flag to stop at first successful response (#571)

xbol0:
# Description

Added `-ss` flag to stop at first successful response.

I think that it should be conflict to other stop flags. So I made it throw a warning message when user provided them.

Fixes: #571 

## Additonally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 - master
   - New
+    - Added `-ss` flag to stop at first successful response
   - Changed
   
 - v1.5.0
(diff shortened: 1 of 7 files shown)


joohoi:
Thanks for the PR, and sorry it took so long to get around to review it. 

I do like the idea for the flag, but I would argue that we should modify the behavior a bit from this initial suggestion. If I'm not mistaken the potential use case example for this is to fuzz things like credentials and such, and for that I believe we should abstract the definition of success a bit. So how about changing the behavior to stop on first match instead of hardcoded status codes etc?

opabravo:
Bump

PR #589, opened 2022-09-21
Feat header file

xbol0:
# Description

Add `-header-file` flag to read headers from file, it's convenient to pass a lot of stationary header params.

I have test it can process FUZZ keyword.

I am considering if it needs header provider to implement instead.

## Additonally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 - master
   - New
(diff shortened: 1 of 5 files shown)


joohoi:
Thanks for the PR! 

I may be missing some use case, so could you enlighten me a bit about a potential situation where using `-request` does not fill this purpose?

PR #595, opened 2022-10-15
fix: changed usage from version 1.3.0 to version 1.5.0

arthur4ires:
Usage was out of date so I switched to the latest version.


joohoi:
Thanks for the PR! Merging.

PR #640, opened 2023-02-07
Fix HTML output

Hazegard:
# Description

Since FFUF 2, the HTML output is broken (`can't evaluate field HTMLColor in type output.htmlResult` ). This PR aims at fixing this bug and adds several improvement regarding the raw results stored in the HTML result

### DIFF
--- a/pkg/output/file_html.go
+++ b/pkg/output/file_html.go
@@ -23,6 +23,8 @@ type htmlResult struct {
 	ResultFile       string
 	Url              string
 	Host             string
+	HTMLColor        string
+	FfufHash         string
 }
 
 type htmlFileOutput struct {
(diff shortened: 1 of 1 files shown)


joohoi (approved):
Thanks for fixing the HTML issue in this PR, lgtm!

HackerTXK:
> # Description
> Since FFUF 2, the HTML output is broken (`can't evaluate field HTMLColor in type output.htmlResult` ). This PR aims at fixing this bug and adds several improvement regarding the raw results stored in the HTML result

Excuse me, Sir. I'm new to information security.I saw in your comments that you fixed this bug, but I still have this problem. I use digitalocean's vps, Ubuntu 22.10
The following is the detailed error content:
**[ERR] template: output.html:74:94: executing "output.html" at &lt; $result.HTMLColor&gt; : can't evaluate field HTMLColor in type output.htmlResult**

By the way, when the output format is html, the result is empty, but the other formats are fine

0xgalal:
Any fix ?

Issue #645, opened 2023-02-12
How to get back the 1 line verbosity like v1.5


C4lputer:
Now it takes 3 lines for 1 result which to me is quite frustrating

![image](https://user-images.githubusercontent.com/93414512/218296059-d04d2ac2-3ca8-41e5-845a-638f1e521b8e.png)

joohoi:
Hi, thanks for opening the issue. This is a bug in case you have no
scrapers enabled. With no scrapers it should default to single-line output,
but the check for it seems to be broken.

However in the case you are using scrapers, the multi-line output is the
only possible output format.

C4lputer:
`ffuf -w /opt/SecLists/Discovery/Web-Content/raft-medium-directories-lowercase.txt -u "http://google.com/FUZZ"`

I just fuzz like I always do, but when I update to version 2 this happened. So I supposed this is a bug and there will be a fix soon?
Thank you for the quick response.

riramar:
Same problem here. AFAIK it's not possible to create filters like:

(HTTP code 400) OR (HTTP code 404) OR (HTTP code 301 AND Words 456)

So I was filtering out using egrep since the results were present in one line.
Now I'm getting 3 lines per result even with no scrapers.

dhtzs:
I'm experiencing the same issue. Seems like ffuf 2.0 now takes 3 lines for a result without using scrapers, which is frustrating indeed.

vionde:
Hope it will be fixed soon, but until that `go install github.com/ffuf/ffuf@v1.5.0`

Anyway, thanks for your work!

0xGreen:
Hi there,

Fuzzing like i used to do (no `-scrapers`)

```bash
ffuf -u http://example.com -w wordlist.txt -H 'Host: FUZZ.example.com' -mc all -fw 1288
```
but the output is really weird, like 3 lines for 1 hit.

Just like @C4lputer, too much for me as well.

![image](https://user-images.githubusercontent.com/6963584/220812022-d071e999-8551-4c56-a7d1-ea8390b75632.png)

p0dalirius:
Hi,

I've fixed the bug, proposed a pull request and wrote details on what was happening in https://github.com/ffuf/ffuf/pull/656

It now works awesomely on my end.

Best regards,

cardassian-tailor:
Just wanted to say I also am having this issue and look forward to this being fixed. Thank you.

0xCGonzalo:
Hi,

[...]

PR #656, opened 2023-03-03
[Fix] Fixed multiline output when using only one fuzz variable (issue #645)

p0dalirius:
# Description

Fixed multi-line output when using only one fuzz variable (issue #645)

Fixes: #645

---

## Bug source

In version 2.0 of fuff, `FFUFHASH` was added. 

https://github.com/ffuf/ffuf/blob/b2c1f9471f5bc275912c9c8194e88593d2636597/CHANGELOG.md?plain=1#L8

This variable is set in the `Input` map of the result structure:

https://github.com/ffuf/ffuf/blob/b2c1f9471f5bc275912c9c8194e88593d2636597/main.go#L397

This triggers two bugs

### Bug 1

Because of this, the size of `res.Input` is now at least 2. Therefore all output modes will be multi-lines since the condition `len(res.Input) > 1` is never met in the function [`PrintResult()`](https://github.com/p0dalirius/ffuf/blob/master/pkg/output/stdout.go#L369) in `pkg/output/stdout.go`.

https://github.com/p0dalirius/ffuf/blob/b2c1f9471f5bc275912c9c8194e88593d2636597/pkg/output/stdout.go#L375-L377

Changing the condition `len(res.Input) > 1` to `len(s.fuzzkeywords) > 1` fixes this bug.

### Bug 2

Another bug then occurs when only one fuzz variable is passed and results are printed. When `prepareInputsOneLine()` in `pkg/output/stdout.go` gets called, the function iterates on res.Input, therefore printing `FFUFHASH` too.

https://github.com/p0dalirius/ffuf/blob/master/pkg/output/stdout.go#L386

https://github.com/p0dalirius/ffuf/blob/master/pkg/output/stdout.go#L395

Changing the iteration on `res.Input` to an iteration on `s.fuzzkeywords` fixes this bug.

---

## New behavior

When fuzzing only one variable:

image

When fuzzing two and more variables:

image

Best regards,

### DIFF
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -382,7 +382,7 @@ func (s *Stdoutput) PrintResult(res ffuf.Result) {
 
 func (s *Stdoutput) prepareInputsOneLine(res ffuf.Result) string {
 	inputs := ""
-	if len(res.Input) > 1 {
+	if len(s.fuzzkeywords) > 1 {
 		for _, k := range s.fuzzkeywords {
 		    if ffuf.StrInSlice(k, s.config.CommandKeywords) {
 				// If we're using external command for input, display the position instead of input
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -372,7 +372,7 @@ func (s *Stdoutput) PrintResult(res ffuf.Result) {
 		s.resultJson(res)
 	case s.config.Quiet:
 		s.resultQuiet(res)
-	case len(res.Input) > 1 || s.config.Verbose || len(s.config.OutputDirectory) > 0 || len(res.ScraperData) > 0:
+	case len(s.fuzzkeywords) > 1 || s.config.Verbose || len(s.config.OutputDirectory) > 0 || len(res.ScraperData) > 0:
 		// Print a multi-line result (when using multiple input keywords and wordlists)
 		s.resultMultiline(res)
 	default:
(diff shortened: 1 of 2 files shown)


p0dalirius:
Hey @joohoi

Can you take a look at this? Everything is fixed in this PR, and the current bug (https://github.com/ffuf/ffuf/issues/645) is really annoying.

Best regards,

pkg/output/stdout.go
```diff
@@ -383,21 +383,21 @@ func (s *Stdoutput) PrintResult(res ffuf.Result) {
 func (s *Stdoutput) prepareInputsOneLine(res ffuf.Result) string {
 	inputs := ""
 	if len(res.Input) > 1 {
```
joohoi, line 385:
```suggestion
	if len(s.fuzzkeywords) > 1 {
```

joohoi: This needs to be changed to observe the fuzzkeywords variable as well.

joohoi (changes requested):
Thanks for the fix. Besides the one comment this LGTM.

joohoi (approved):
Commited the suggestion, so this is ready to merge.

sk0x1234:
using the latest git build

silent/quit ` -s` option  producing in wrong format
FUZZ anf FFUFHASH interchanged. im using latest git build

image

image

felipecaon:
Seems like this issue is back, using `v2.0.0`

Issue #660, opened 2023-03-28
Output format columns do not line up properly


robertstrom:
I have tried outputting all formats (and just csv) using a command like the one shown below:

```
ffuf -c -w /usr/share/wordlists/seclists/Discovery/Web-Content/big.txt -u http://www.holo.live/FUZZ -t 25 -fc 404 -e .php,.html,.htm,.txt,.bak,.zip,~ -o ffuf-www.holo.live-dirbust -of all
```
It does not seem to matter what format the output is in, the data in each column is incorrect. An example of the markdown output is shown below

```
| FUZZ | URL | Redirectlocation | Position | Status Code | Content Length | Content Words | Content Lines | Content Type | Duration | ResultFile | ScraperData
  | :- | :-- | :--------------- | :---- | :------- | :---------- | :------------- | :------------ | :--------- | :----------- | :------------ |
  | 4500679 | .htaccess | http://admin.holo.live/.htaccess |  | 121 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 170.385517ms |  |  |
  | 450067a | .htaccess.php | http://admin.holo.live/.htaccess.php |  | 122 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 170.991239ms |  |  |
  | 450067b | .htaccess.html | http://admin.holo.live/.htaccess.html |  | 123 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 170.689087ms |  |  |
  | 450067c | .htaccess.htm | http://admin.holo.live/.htaccess.htm |  | 124 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 171.412565ms |  |  |
  | 450067d | .htaccess.txt | http://admin.holo.live/.htaccess.txt |  | 125 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 171.058853ms |  |  |
  | 450067e | .htaccess.bak | http://admin.holo.live/.htaccess.bak |  | 126 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 170.998954ms |  |  |
  | 450067f | .htaccess.zip | http://admin.holo.live/.htaccess.zip |  | 127 | 403 | 280 | 20 | 10 | text/html; charset=iso-8859-1 | 169.575039ms |  |  |
[...]

PR #672, opened 2023-04-21
Fix md output format

joohoi:
Fixes: #660


sk0x1234:
need to release the updated package after this push,As of now current previous version output is unusable.

PR #677, opened 2023-05-10
Don't write a finishing newline in silent mode

Nirusu:
# Description
When running ffuf in silent mode, a new line is still written on every line at the end of a run of ffuf.
While this looks fine if there are any matches, it will just spam empty new lines whenever:

- No matches are found
- If you redirect stdout but keep stderr on console

Piping ffuf multiple times with different targets, this approx. looks like this:

https://github.com/ffuf/ffuf/assets/26608194/6194f5ec-8e51-496a-9149-68b04cc87658

Thus, let's remove printing the new line when in silent mode.

## Additonally

-> I skipped since I thought this change is too insignificant.

### DIFF
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -307,7 +307,9 @@ func (s *Stdoutput) Finalize() error {
 			s.Error(err.Error())
 		}
 	}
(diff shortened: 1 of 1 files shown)


joohoi:
Thanks for the PR, merging

Issue #691, opened 2023-06-14
Recursion Status


aristosMiliaressis:
Hi,

i would like an option to control what status codes cause recursion to happen

-rs, --recurion-status 300-308,401,403

PS: i am planning on implementing this feature myself, just opening this issue as a heads up and to receive feedback while i'm working on this and the other issues i opened

PR #693, opened 2023-06-15
Recursion status

aristosMiliaressis:
# Description

I added the option `-recursion-status` to control what response codes cause recursion to happen, the main use case is to recurse on 403.

Also i added the option `-acp` to auto calibrate per path, so if a 403 path is recursed and and all subsequest requests also receive 403 they are not also recursed.

Fixes: #691

### DIFF
--- a/pkg/ffuf/job.go
+++ b/pkg/ffuf/job.go
@@ -468,7 +468,7 @@ func (j *Job) runTask(input map[string][]byte, position int, retried bool) {
 		}
 	}
 
-	if j.Config.Recursion && j.Config.RecursionStrategy == "default" && len(resp.GetRedirectLocation(false)) > 0 {
+	if j.Config.Recursion && j.Config.RecursionStrategy == "default" && intSliceContains(j.Config.RecursionStatus, int(resp.StatusCode)) {
 		j.handleDefaultRecursionJob(resp)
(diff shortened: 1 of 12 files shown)


pkg/ffuf/job.go
```diff
@@ -499,7 +505,7 @@ func (j *Job) handleGreedyRecursionJob(resp Response) {
 // not been reached
 func (j *Job) handleDefaultRecursionJob(resp Response) {
 	recUrl := resp.Request.Url + "/" + "FUZZ"
-	if (resp.Request.Url + "/") != resp.GetRedirectLocation(true) {
+	if (resp.Request.Url+"/") != resp.GetRedirectLocation(true) && resp.StatusCode >= 300 && resp.StatusCode < 400 {
```
joohoi, line 508: I think we should honor the values in the new cli parameter for recursion-status here instead of hardcoded values between 300 and 400.

joohoi (changes requested):
Thanks for the PR!

After initial review there was just this one change request that I spotted.

pkg/ffuf/job.go
aristosMiliaressis: nice catch,  i pushed the suggested change

PR #695, opened 2023-06-16
[FEATURE] Automatic ratelimit adjustments

aristosMiliaressis:
# Description

Adds the option `-ar` to enable automatic adjustments to rate limiting based on the amount of 429 received.

i used the following script to test it
```python
#!/usr/bin/env python
# encoding: utf-8
import time
from flask import Flask, redirect

app = Flask(__name__)
clients = {}

@app.route('/admin', methods=['GET'])
def admin():
	return redirect("/admin/", code=302)

@app.route('/admin/<path>', methods=['GET'])
def nested_admin(path):
	id = str(int(time.time()))
	if id in clients.keys():
		clients[id] += 1
	else:
		clients[id] = 1

	if clients[id] > 10:
		return "Too Fast", 429

	return "Not Found", 404

app.run(debug=True)
```

and the requests per second went from 120 to 9 when it started on the /admin/ path
Screenshot 2023-06-16 103600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 - master
   - New
+    - New command line parameter for auto adjasting ratelimiting based on amount of 429 `-ar`
   - Changed
     - Explicitly allow TLS1.0 
     - Fix markdown output file format
--- a/help.go
+++ b/help.go
@@ -61,7 +61,7 @@ func Usage() {
 		Description:   "",
 		Flags:         make([]UsageFlag, 0),
 		Hidden:        false,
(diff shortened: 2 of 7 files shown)


joohoi:
Thanks for the PR! 

This looks good, but I have some additional ideas that could make it awesome. Some of them may require some larger changes though:

- In addition to looking into HTTP response status codes, there is a [IETF draft (to become RFC) for standardized rate limit header](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html) , and `X-Rate-Limit` that is currently used in the wild as a standards compliant (X- for non-standard) header that we could use to "predict" the exact rate limit and play according to that.
- We could create a "to-retry-later" cache of payloads that got the `429` response, to be tried after the wordlist(s) have been exhausted. For this I suggest using an additional structure, or even extending the `InputProvider` interface, to not to try to add those entries back to the InputProvider (this gets really hairy really fast, especially when dealing with multiple wordlists algorithm we use atm).

PR #697, opened 2023-06-25
Disable HTTP/2 support when not enabled

alexvec:
# Description

The parameter `-usehttp2` is disabled by default, however the ffuf client doesn't actually disable HTTP/2 requests when that parameter is set to `false`.

As it is [officially documented](https://pkg.go.dev/net/http#pkg-overview)

`Starting with Go 1.6, the http package has transparent support for the HTTP/2 protocol when using HTTPS. Programs that must disable HTTP/2 can do so by setting Transport.TLSNextProto (for clients) or Server.TLSNextProto (for servers) to a non-nil, empty map.`

However, this isn't being done in the runner client, https://github.com/ffuf/ffuf/blob/5fd821c17d5589c8f66211389d99e6f0885354d1/pkg/runner/simple.go#L51

Not forcing HTTP/2 is not the same as disabling HTTP/2.

I have modified the client code to disable HTTP/2 requests completely if the parameter `-usehttp2` , by setting an empty map under `TLSNextProto`:
`transport.TLSNextProto = map[string]func(string, *tls.Conn) http.RoundTripper{}`

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,6 +2,7 @@
 
 * [adamtlangley](https://github.com/adamtlangley)
 * [adilsoybali](https://github.com/adilsoybali)
+* [alexvec](https://github.com/alexvec)
 * [AverageSecurityGuy](https://github.com/averagesecurityguy)
 * [bp0](https://github.com/bp0lr)
 * [bjhulst](https://github.com/bjhulst)
--- a/pkg/runner/simple.go
+++ b/pkg/runner/simple.go
@@ -44,10 +44,7 @@ func NewSimpleRunner(conf *ffuf.Config, replay bool) ffuf.RunnerProvider {
 		}
 	}
 	simplerunner.config = conf
-	simplerunner.client = &http.Client{
-		CheckRedirect: func(req *http.Request, via []*http.Request) error { return http.ErrUseLastResponse },
-		Timeout:       time.Duration(time.Duration(conf.Timeout) * time.Second),
-		Transport: &http.Transport{
+	transport := &http.Transport{
 			ForceAttemptHTTP2:   conf.Http2,
 			Proxy:               proxyURL,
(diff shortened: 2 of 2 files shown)


joohoi:
I think we should maintain sensible defaults, and to me opportunistic use of HTTP/2 would be a sensible default. I would love to see the behavior being:
 - Default: opportunistic use of HTTP/2
 - A flag to force HTTP/2 (I would assume rare, but still realistic use case), like `-usehttp2` is atm
 - A flag to _disable_ HTTP/2, something like `-nohttp2`

What do you think? And would you be up to making the change?

alexvec:
Sure, I like the idea. I will make the changes and update it.

alexvec:
Hi, sorry for the delay.

I have a proposed idea which might make for a better implementation:

As it currently stands, `http2` flag will enable HTTP/2 to make opportunistic use of HTTP/2. At the moment, by default that flag is set to `false`, which means that HTTP/2 is currently not being used for any request under the default behaviour.

There are several options:
1. Introduce default behaviour to use HTTP/2 opportunistically, if the server provides support for it, otherwise it would fallback to HTTP/1. Therefore, we can remove `http2` flag, add a new flag called `nohttp2` which is set to `false` by default. When said flag is set to `true` will mean all requests are never going to be handled by HTTP/2, rather by HTTP/1. (This is in fact, what currently happens when running `ffuf` under the default settings.
2. Leave default behaviour as is, HTTP/2 will not be used, unless `http2` flag is set. (At the moment, this works, however official Go documentation also mentions to set TLSNextProto to a non-nil, empty map, which is a small change that can be introduced).
3. Do option 1, and moreover, introduce a flag `forcehttp2` which forces exclusive use of HTTP/2 requests, or else errors.

For the time being, I consider option 1 to be the simplest, and I think enabling HTTP/2 opportunistically as a default behaviour is a good idea. Nevertheless, please let me know what you think

PR #705, opened 2023-08-02
Do not add request time to delay from -p flag

codesoap:
# Description
Right now the delay from the `-p` is added to the time that a request already consumes. I think this is less useful than interpreting the delay as the time between the beginning of each request. This is what I implemented here.

My use-case is, that I'm fuzzing HTTP paths on a target, but I've been asked to make no more than two requests per second. Thus I tried to use `ffuf -t 1 -p 0.5 ...`, but noticed, that I just got 1 req/seq instead of 2.


joohoi:
Hi, thanks for the PR. 

While I certainly appreciate the work gone towards creating this. I however think that we should leave the behavior as-is, as per definition the `-p` flag states:

"Seconds of `delay` *between* requests"

To remediate your use case there already is a rate limiting flag, and  `-rate 2` should just do the job for you.

codesoap:
Alright, thanks. Honestly I hadn't seen the `-rate` option previously, as I had used an older version of `ffuf`. This flag does indeed satisfy the requirements I described.

Slightly off-topic question: Do you plan on supporting floats with the `-rate` option in the future? I once already had the case were I was asked to limit my requests to 100 per minute. This would translate to `-rate 1.66`, which is not allowed right now.

joohoi:
I doubt floats will be coming to rate, unfortunately. I bet the target would be cool with 120 req/min (2 per second) though.

PR #709, opened 2023-08-14
Edit outputDirectory

alanEG:
# Description

The current outputDirectory function creates unreadable content that makes it difficult for anyone to handle the saved content. This is because there is no indication of where the content data starts. To address this, I've implemented a new saving method. In this approach, I am saving the request headers and response headers in JSON format on the first line of the file. Following that, the response content. This should help make the output more organized and easier to handle.

Fixes: #(issue number)

## Additonally
....

### DIFF
--- a/pkg/ffuf/request.go
+++ b/pkg/ffuf/request.go
@@ -16,6 +16,14 @@ type Request struct {
 	Raw      string
 }
 
+type RequestWithoutExcludedFields struct {
+	Method  string
+	Host    string
+	Url     string
+	Headers map[string]string
+	Data    []byte
+}
+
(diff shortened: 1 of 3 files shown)


joohoi:
Hi, thanks for the PR.

While I do appreciate the time put in to writing this, I need to kindly decline from merging it.
The current output format is in the raw http response as defined in HTTP standard [RFC2616](https://datatracker.ietf.org/doc/html/rfc2616#section-6) and there are parsers for http responses available in all of the widely spread programming languages. 

In addition to this, there's a good bunch of automation and implementations people have built on top of the current behavior, and without a _very_ strong case, I'm not willing to break that.

PR #728, opened 2023-09-19
Add alias for interactive commands

p1gp1g:
# Description

Many interative tools have aliases, queueshow, queuedel, and queueskip are often used. Having those aliases are a lot faster :)

## Additonally

- [ ] If this is the first time you are contributing to ffuf, add your name to `CONTRIBUTORS.md`.  #727 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 - master
   - New
+  - Add `q`, `qd`, `qs` alias for `queueshow`, `queuedel` and `queueskip`
   - Changed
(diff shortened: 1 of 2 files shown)


joohoi (changes requested):
Thanks for the PR!

This looks good, but it needs a change to help text (and `README.md` accordingly). I'd prefer the help text to only show the shortcuts that this PR adds, while still keeping the long form commands "hidden" in the background. I think backwards compatibility is a good thing.

p1gp1g:
Done :)

Issue #525, opened 2022-02-25
Retrieve CSRF token before each request


chcore:
Hello,

Let's say my POST request needs a username, password and a CSRF token. The username/password can be fed from a text file, but the CSRF token must be retrieved at runtime, right before making the request (for example, by doing a CURL request and extracting the value with regex).

1. Is this possible with ffuf?
2. If so, how does it work with the "cluster" mode, where every combination of username/password must be tested?
3. Is it possible to have multiple dynamically generated values (say there are 2 CSRF tokens)?

Many thanks

bsysop:
Hi @chcore, not actually! You can do it using a bash script and then sending one request with FFUF, but honestly it don't makes sense, ffuf is powerful to send multiple requests.

I recommend you to do it with **Burp Intruder** 👀 

Happy hacking!

chcore:
Thank you for your answer @bsysop,

I agree that it wouldn't make sense to execute just one request at a time with ffuf. I was hoping with my question that I would find in ffuf an efficient CLI alternative to Burp Intruder, at least for my simple needs.

Thanks again

bsysop:
That would be great, probably in the future, but it's pretty hard to implement all that considering just the CLI
FFUF is not that kind of "easy" tool to use, more features means more complexity, which is not bad necessarily 👀 .

Happy hacking!

bsysop:
_Adding some comments to this, for the future possible implementation :)_

The setup for this its a bit complex, a bit hard to explain how to use (you need to debug before really using it) and for sure not easy to maintain, but for sure is interesting.

**SQLMAP**
- It allow you to get the CSRF token from another URL
- `--csrf-token` don't works with `--threads`, so it limit the concurrency (but I think they could fix this at some point)
> SQLMap can handles cookies too, it capture from the original page and send to the second (if you want), but its far away from perfect.

**BURP** (PRO only maybe)
- You can configure MACROS (amazing feature btw), allowing to "capture" any string (using REGEX) from **page A** to send to **target B**.

**Implementation suggestion:**
Sometimes you could need more than just a CSRF token.. it could be cookies or a simple string to use in the headers, so probably the best implementation could be something like:

```
[...]
-preflight-request https://target.tld/ -capture-regex 'your_string_extraction_regex_here':REGEX -H "Header: REGEX"
[...]
```

That way you could use for CSRF tokens, Cookies, headers, anything you want.

PR #845, opened 2025-04-25
So-called "preflight requests" for getting csrf and cookies, add pausing on some responce codes (as stop on 403 but pausing...)

virusvfv:
First interesting feature is so-called pre-flight requests as discussed in #525
So it can be possible to get CSRF tokens and cookies before each requests
Example:
```
ffuf -c -w /temp/wordlist.txt:FUZZ -u https://ffuf.io.fi/login 
     -preflight-request https://ffuf.io.fi/ 
     -capture-regex 'Set-Cookie: BruteCookie=(.*); SameSite':REGEX1
     -capture-regex 'csrftoken=(.*) blablahtml':REGEX2
     -H "Cookie: BruteCookie=REGEX1"
     -H "Foo: bar"
     -preflight-header "Cookie: BruteCookie=Empty"
     -preflight-header "FirstVisit: 1"
     -d "csrf=REGEX2&login=user&password=FUZZ"
```

Second feature is pausing on specified codes. For example if ffuf got 429 of  403 from WAF it can pause for some time and continue after WAF releasing. 
Ex:
`
./ffuf -c -w /temp/wordlist.txt:FUZZ -u https://ffuf.io.fi/FUZZ -pausecode 403,429 -pausetime 60,90,120,180,300
`

So first pause will be 60 sec, second - 90 sec, etc. Fifth, six and other pauses  will be 300 seconds.


ManU4kym (approved):

cheshie:
Great feature! I used it recently on an engagement - had to fetch token (tied to session) from JSON embedded in login page before sending it to brute-force the login. This is how I used it: 
```
./ffuf -preflight-request http://test.me/login -request ~/login.req.txt -w ~/users.txt:FUZZ1 -w ~/passw.txt:FUZZ2 -capture-regex '"token":"(.*)"':REGEX1 -p 0.1-1 -capture-regex 'Set-Cookie: sessid=(.*); path=/; secure; HttpOnly':REGEX2 -v
```

Whereas login.req.txt looks like: 
```
POST /login HTTP/1.1
Host: test.me
Cookie: sessid=REGEX2;
Content-Type: application/x-www-form-urlencoded


```

Unfortunately ffuf was copying REGEX to headers in payload sent which gave me duplicated header. Had to put this if inside for loop and it worked (pkg/runner/simple.go line 154):
```
	for k, v := range req.Headers {
		if !strings.Contains(v, "REGEX"){
		  httpreq.Header.Set(k, v)
	  }
	}
```
Hope it helps somebody, and please merge this, it's very useful option and I'll definitely make use of it in the future!

PR #887, opened 2026-03-25
Payload tampers

Brum3ns:
# Description

Payload tampers now added to Ffuf using [yaegi](https://github.com/traefik/yaegi). Users can now easily write Go interpreter tampers, avoiding compiling and version / dependency mismatch making it fast to develop and add. I already added 34 tampers with the help of [YesWeHack](https://yeswehack.com/) to the community and hopefully more will come. Enjoy!

## Development
I made it possible to write the tampers in a very simple structure. Just add a description and the execution function itself.

**Example template:**
```go
package tamper

type T struct{}

func (t T) Desc() string {
	return "Description of what the tamper do with some example input -> output."
}

func (t T) Exec(payload string) string {
  // Code ... (whatever you want to do with the payload
  return payload
}

var Tamper T
```

```go
package tamper

type T struct{}

func (t T) Desc() string {
	return "add prefix \"yeswehack_\""
}

func (t T) Exec(payload string) string {
	return "yeswehack_" + payload
}

var Tamper T
```

> comment4parenthese
```go
package tamper

type T struct{}

func (t T) Desc() string {
	return "comment before parentheses (Ex: alert(...) => alert/**/(...)"
}

func (t T) Exec(payload string) string {
	buf := make([]byte, 0, len(payload)*2)
	for i := 0; i < len(payload); i++ {
		if payload[i] == '(' {
			buf = append(buf, '/', '*', '*', '/')
		}
		buf = append(buf, payload[i])
	}
	return string(buf)
}

var Tamper T
```

## Teasers:
**Downloading payload tampers:**
image

**List tampers:**
image

## Additonally

Cheers,
Brumens


joohoi:
Hi, thanks for this! And thanks to YesWeHack too. The tamper collection itself is great and something I'd like to have in ffuf.

Before going through the code in detail, one design point that should steer the implementation:

ffuf already has encoders (`-enc`) doing basically the same thing as tampers; both are just per-keyword payload transformations applied at the same spot. This PR adds a second, parallel mechanism alongside them, with the dispatch logic mostly copy-pasted from the encoder path. I'd rather we merge the two into a single transformation path than maintain two that overlap; they should share the plumbing, not duplicate it.

Couple of related points to sort out as part of that:

  - Speed: encoders are compiled, while tampers run through a Go interpreter (yaegi) on every payload. That's a significant difference on the hottest path in the tool, so the built-in transforms should stay compiled rather than interpreted.
  - Downloading: pulling .go files and executing them isn't something I want to ship as-is. If the download feature stays, it needs a fixed, hardcoded source (no user-provided URL). It's not too hard to imagine an offensive AI workflow going awry and executing some unintended code due to this.

Happy to go deeper on any of these. Keen to get it in, just in a shape that fits how ffuf already works :)

What do you think, how should we proceed?

brumensywh:
Hey Joona,

You mentioned that Yaegi is used to interpret tampers, and that:

> "...the built-in transforms should stay compiled rather than interpreted."

While Yaegi is indeed a Go interpreter, it supports a hybrid approach. For example, when a tamper calls a function such as `strings.Replace(...)`, Yaegi use the compiled Go implementation already available in memory. This means much of the underlying work still runs as compiled code, providing performance close to native execution.

Maintaining a fully compiled tamper system would also be difficult, particularly because each Ffuf update could introduce dependency hell especially for the community if they write tampers. Yaegi provides a practical middle ground: the tamper logic remains flexible and interpreted, while calls to registered Go packages and functions can use their compiled implementations. In practice, the performance difference should therefore be minimal.

To confirm that I understand your dev proposal correctly:

* Remove the separate encoder feature and move all encoders into the tamper system, providing a single feature for payload modification.
* Use a hardcoded URL so that tampers can only be fetched from the official Ffuf GitHub repository.


Sounds good?

joohoi:
Thanks for digging into this. You're correct that yaegi calls into the registered stdlib through reflection, so something like `strings.Replace` does run the compiled implementation rather than an interpreted one. The part that still runs interpreted is the tamper's own function body, which for most of these is a loop over the payload bytes, and that loop is what runs on every payload we send. For a one-liner like `"yeswehack_" + payload` the interpreted portion is small enough that it does not matter, but several of the tampers you added (`space2comment`, `comment4parenthese`, and similar) are manual byte loops whose interpreted cost scales with payload length, and that is exactly the path that gets hit millions of times in a single run, so I do not think we can wave the performance question away without measuring it.

The reason I do not think we need to measure it, though, is that the dependency and version-skew argument you are making only applies to tampers that a user writes themselves outside the tree. The thirty-four you are shipping are not in that category, since they live in the repository and get released with ffuf, so there is no version mismatch to protect against for them and they can simply be compiled.

There is also a robustness problem with the interpreted approach as it stands. `PayloadTamper.Execute` calls `t.Exec(payload)` in a loop with no `recover()` around it, and it runs on the per-payload path, so if an interpreted tamper panics on some input, whether from a bug in the tamper or from input it did not expect, the panic propagates up and takes down the whole ffuf process in the middle of a scan. The `Exec(payload string) string` signature makes this worse, because a tamper that hits a bad state has no way to return an error, so its only options are to return wrong output or to panic. A compiled encoder in pencode returns `([]byte, error)`, and the chain already checks that error on each step, so a failing transform surfaces as an error on that request instead of crashing the run.

That connects to the larger structural point. The encoders in ffuf are already a single transformation path, and they already live in a separate reusable library, `ffuf/pencode`, applied in the input path as one `chain.Encode()` call. Rather than adding a second mechanism inside ffuf, the tampers fit as additional transforms in that same chain. I ported four of yours into pencode as compiled encoders to check that this actually works, including `comment4parenthese` since it was the loop-heavy one, and put it on a branch so it is easy to look at in context: https://github.com/joohoi/pencode/tree/tamper-prototype (the whole change is [`pkg/pencode/tampers.go`](https://github.com/joohoi/pencode/blob/tamper-prototype/pkg/pencode/tampers.go) plus a test that runs them through the real `pencode.Chain`). Each one is a small struct implementing the existing `Encoder` interface, and they self-register, so adding a tamper is one new file plus one line and nothing in ffuf itself has to change:

```go
const tamperType = "tampers"

func init() {
	availableEncoders["space2plus"] = Space2Plus{}
	availableEncoders["space2comment"] = Space2Comment{}
	availableEncoders["comment4parenthese"] = Comment4Parenthese{}
	availableEncoders["reversepayload"] = ReversePayload{}
}

// Space2Comment replaces spaces with an SQL empty comment "/**/".
type Space2Comment struct{}

func (Space2Comment) Encode(in []byte) ([]byte, error) {
	out := make([]byte, 0, len(in)*2)
	for _, c := range in {
		if c == ' ' {
			out = append(out, '/', '*', '*', '/')
		} else {
			out = append(out, c)
		}
	}
	return out, nil
}
func (Space2Comment) HelpText() string { return "tamper: replace space with multi-line comment (/**/)" }
func (Space2Comment) Type() string     { return tamperType }

// Comment4Parenthese inserts "/**/" before every '(', the loop-heavy example from above.
type Comment4Parenthese struct{}

func (Comment4Parenthese) Encode(in []byte) ([]byte, error) {
	out := make([]byte, 0, len(in)*2)
	for _, c := range in {
		if c == '(' {
			out = append(out, '/', '*', '*', '/')
		}
		out = append(out, c)
	}
	return out, nil
}
func (Comment4Parenthese) HelpText() string {
	return "tamper: comment before parentheses, e.g. alert(...) => alert/**/(...)"
}
func (Comment4Parenthese) Type() string { return tamperType }
```

With that in place, `-enc FUZZ:space2comment` works through the existing plumbing with no interpreter involved, and the tampers compose with the encoders that are already there, so a chain like `space2comment` followed by `urlencode` runs in one pass. I operated on `[]byte` directly rather than `string` on purpose, since that is pencode's native type and it avoids the round trip and the binary-payload corruption that the `Exec(string) string` version has.

To answer your two questions directly. On merging the two features, I agree there should be one path, but the direction I have in mind is folding the tampers into the existing compiled encoder chain rather than moving the encoders into a yaegi based system, so the encoders stay compiled and the tampers join them as compiled transforms. Adding a new one stays a one file change, which keeps the contribution workflow you wanted, with the difference that it ships in a release instead of being fetched at runtime.

One thing worth deciding separately from the plumbing is the flag name. Mechanically a tamper and an encoder are the same operation, a per-keyword transform applied to the payload, but from a user's point of view `-enc FUZZ:space2comment` reads oddly, because they are thinking of it as a tamper rather than as an encoding step, and the two words carry different intent in this context even though the implementation is identical. Since both would feed the same chain, we can keep the single code path and still expose a name that matches intent by adding `-tamper` as an alias for `-enc` that populates the same encoder chain, so `-tamper FUZZ:space2comment` and `-enc FUZZ:urlencode` both end up in the same pipeline and can be combined freely. The prototype already tags these transforms with `Type() == "tampers"`, so the help output and a listing command can group them under a tampers heading separately from the encoders while they share the underlying machinery, which gives users the mental model they expect without us maintaining two implementations.

On the download feature, I would rather remove it than hardcode the URL, and the reasons are in the current code. The download URL is user settable through `-download-tampers-url` (the default points at the ffuf repository, but a user, a config file, or some automated workflow can point it anywhere), and the loader calls `i.Use(stdlib.Symbols)`, which means an interpreted tamper has the entire standard library available to it, including `os/exec`, networking, and filesystem access, rather than being limited to string manipulation. There is also no signature or checksum check on what gets downloaded. Taken together that is fetching arbitrary code from a URL and running it with the process privileges, out of band from the release that a user actually reviewed and pinned. Hardcoding the URL narrows the first part but does not change the fact that a tamper is unsandboxed code that is not tied to a reviewed release, so I would rather drop the runtime download path entirely. If the built in tampers are compiled and new ones come in through pull requests, the downloader is not needed for anything.

I do think yaegi has a place, just a narrower one, which is for tampers a user writes locally and wants to run without setting up a Go toolchain and rebuilding ffuf. That would be opt in, loaded only from a local directory the user points at, and documented clearly as running arbitrary code with ffuf's privileges so people only load tampers they trust. It just should not be the path for the shipped tampers, and it should not be fed by a downloader.

To make that local path concrete, here is how it would work end to end for someone who wants a transform that ffuf does not ship. They write a single Go file in the format the interpreter expects, which is the same format the current PR already uses:

```go
package tamper

type T struct{}

func (t T) Desc() string {
	return "prefix the payload with a fixed marker"
}

func (t T) Exec(payload string) string {
	return "myprefix_" + payload
}

var Tamper T
```

They save it in the tampers directory, where the filename without the extension becomes the tamper name, so this file would live at `~/.config/ffuf/tampers/myprefix.go` on Linux, which is the same XDG config location the PR already defaults to and which can be overridden with `-tampers-dir`. Then they run it by name through the same flag as the built in transforms:

```
ffuf -w wordlist.txt -u https://target/FUZZ -tamper FUZZ:myprefix
```

When ffuf resolves `myprefix`, it does not find it in the compiled registry, so it falls back to looking for `myprefix.go` in the tampers directory and interprets it with yaegi. Because that is running code the user placed on disk, this fallback is opt in and prints a warning that interpreted tampers run with ffuf's privileges, so only tampers the user trusts should be placed there. The built in tampers never reach this path, since they resolve from the compiled registry first, so the interpreter is only ever involved for a user's own local files.

The end result is actually less code than the current PR, since `download.go` goes away and yaegi comes off the hot path, and the tamper collection itself, which is the valuable part of this and something I want in ffuf, stays. The branch above is a starting point if you want a concrete target to port against, and I am happy to turn it into a proper pencode PR we iterate on together. What do you think?

PR #893, opened 2026-05-01
Detect terminal width and shorten progress line on narrow terminals

### LINKED ISSUE #166: Disable bunch of errors in CLI

ChrisJr404:
The progress line written to stderr in `Stdoutput.Progress` is ~100 characters wide. On terminals narrower than that, it wraps onto the next line and the next refresh's `\r` returns the cursor to the start of the wrapped line instead of the start of the original line, producing the corrupted output reported in #166:

```
:: Progress: [646/6245] :: Job [1/1] :: 80 req/sec :: Duration: [0:00:08] :: Err:: Progress: [675/6245] :: Job [1/1] :: 84 req/sec :: Duration: [0:00:08] :: Err...
```

This detects the terminal width via `ioctl(TIOCGWINSZ)` on POSIX and `GetConsoleScreenBufferInfo` on Windows (both via the existing `golang.org/x/sys` dependency) and picks the widest of three progressively shorter progress line variants that fits:

1. `:: Progress: [N/T] :: Job [Q/QT] :: R req/sec :: Duration: [h:mm:ss] :: Errors: E ::` (full)
2. `:: [N/T] R r/s h:mm:ss Errs:E` (medium)
3. `:: [N/T]` (tiny — truncated to width if even this is too long)

When the terminal width cannot be determined (stderr is not a tty, or detection fails), the full variant is used so output to pipes and files is unchanged. Existing behavior for `-s` / quiet mode is unchanged.

The progress formatting was extracted into a `formatProgress(status, width)` helper so it can be unit-tested at arbitrary widths without a real tty. The new tests verify that no variant ever exceeds the requested width and that the full variant retains the original content.

Fixes #166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     - Fix a bug in autocalibration strategy merging, when two files have the same strategy key
     - Fix a bug in -or, causing output to not to be written in any case
     - Fix panic when setting rate to 0 in the interactive console
+    - Detect terminal width and render a shorter progress line when the full one would not fit, fixing the wrap-then-overwrite corruption on narrow terminals
   
 - v2.1.0
   - New
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -167,6 +167,14 @@ func (s *Stdoutput) Progress(status ffuf.Progress) {
 		return
 	}
 
+	fmt.Fprintf(os.Stderr, "%s%s", TERMINAL_CLEAR_LINE, formatProgress(status, terminalWidth()))
+}
+
+// formatProgress returns the progress line for the given status, picking the
+// widest variant that fits within width columns. When width <= 0 (e.g. stderr
+// is not a tty, or detection failed) the full variant is returned so output to
+// pipes/files is unchanged.
+func formatProgress(status ffuf.Progress, width int) string {
 	dur := time.Since(status.StartedAt)
 	runningSecs := int(dur / time.Second)
 	var reqRate int64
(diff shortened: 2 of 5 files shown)


joohoi (commented):
Took a proper look at this. The approach is right: measuring the rendered line and shrinking it to fit is the correct fix for #166, and pulling `formatProgress` out as a pure function is what makes it testable without a real tty. Wordlist size is handled correctly too, since you measure the actual string rather than a fixed template, so a huge `-w` just means more digits in `[N/T]` and the fit logic adapts on its own.

A few things before this goes in:

**Windows: shrinking lines will leave residue.** `TERMINAL_CLEAR_LINE` is `"\r\r"` on Windows (no `\x1b[2K`), so a refresh relies on the new line fully covering the old one. The current progress line only ever grows, so that always held. This change can make the line shrink mid-run (terminal resize, a full to medium variant switch, or req/sec dropping a digit), and on Windows the tail of the previous longer line stays on screen. So we'd be trading the wrap corruption for trailing garbage in the same narrow-terminal case we're trying to fix. Cheap fix: right-pad the chosen variant with spaces out to `width` so each frame overwrites cleanly whether or not the terminal can erase to EOL.

**mintty / Git-Bash / Cygwin won't get the fix.** Under those, stderr is a pipe rather than a console, so `GetConsoleScreenBufferInfo` fails and we fall back to the full line, so #166 stays for those users. It degrades gracefully so it's not a blocker, but that's a large chunk of Windows users; worth a line in the changelog so nobody expects it to engage there.

**`len(full) <= width` allows a line exactly `width` columns wide.** Most terminals defer the wrap at the last column so this is fine in practice, but `< width` removes the ambiguity on older or misbehaving ones. Minor, your call.

**go.mod:** `golang.org/x/sys` is currently `// indirect`, but this imports it directly now. Not a build break (it's already in the graph at v0.5.0), but run `go mod tidy` so the annotation is honest.

**Tests:** the width / no-overflow property is a good one to have. Two gaps: nothing pins which variant is chosen at a given width (returning `tiny` unconditionally would still pass), and there's no large-`ReqTotal` case even though that's the interesting one for line length. Worth adding a case with an 8 or 9 digit total to lock the wordlist-length behavior.

None of this is structural, the design is sound. The Windows padding is the one I'd actually want fixed; the rest are nits.

PR #894, opened 2026-05-02
Add --redirect-chain to capture each hop when following redirects

### LINKED ISSUE #324: Follow redirects while printing

ChrisJr404:
Closes #324.

That issue has been open since 2021 with multiple users repeatedly asking for it (most recent ping in 2024-12). Today the behavior is binary: either don't follow redirects (and see only the first 3xx) or follow them (and see only the terminal 200), never both.

This adds a new flag — `--redirect-chain` — that captures each intermediate hop (the URL of the request that returned the 3xx, the status code, and the `Location` header) and surfaces the chain everywhere a result is rendered.

### Behavior

- **Off by default.** No change for anyone who doesn't opt in.
- **`--redirect-chain` implies `-r`** — you have to follow the redirects to capture them.
- **Existing `-r` is unchanged.** When `-r` is set without `--redirect-chain`, the runner takes the same code path as before and `Redirects` is empty.
- **10-hop ceiling**, matching `net/http`'s default, prevents runaway loops.

### Output surfaces

- **Verbose multiline stdout**: one extra `| --> | <status> <location>` line per hop alongside the existing `| URL |` line.
  ```
  [Status: 200, Size: 11, Words: 1, Lines: 1, Duration: 0ms]
  | URL | http://target/admin
  | --> | 301 /login
  | --> | 302 /dashboard
      * FUZZ: admin
  ```
- **JSON**: new `"redirects": [{"url","status_code","location"}]` array per result (`omitempty`, so unchanged when the chain is empty).
- **CSV / MD / HTML**: new `redirects` / `Redirects` column (empty cell when no chain — existing pipelines that count columns still parse). HTML cell content is `html.EscapeString`'d since URLs/Locations originate from untrusted servers.
- **Audit log**: picks up the field automatically via the embedded `Result`.

### Implementation

A per-`Execute` `*[]RedirectHop` is plumbed through the request context. The shared `CheckRedirect` callback set on the `http.Client` reads the slice pointer from `req.Context()` and appends one hop per redirect, using `req.Response` (the 3xx that triggered the redirect) and `via[len(via)-1]` (the request that received it). The struct-typed context key (`redirectChainCtxKey{}`) avoids collisions with other context values.

### Tests

New file `pkg/runner/simple_test.go` covers four cases against an in-process `httptest` server:
1. **`TestRedirectChain_CapturedAcrossThreeHops`** — verifies all hops in a 3-hop chain are captured with correct URL, status, and Location.
2. **`TestRedirectChain_DisabledByDefault`** — `--redirect-chain` off → same as today's default (no follow, empty `Redirects`).
3. **`TestRedirectChain_FollowRedirectsWithoutChainStillFollows`** — `-r` alone still follows but doesn't capture (back-compat lock).
4. **`TestRedirectChain_StopsAfterTenHops`** — 30-hop loop triggers the 10-hop cap with a clean error.

`pkg/output/file_csv_test.go` extends `TestToCSV` for the new empty column and adds `TestToCSV_WithRedirectChain` for the populated case. `pkg/output/audit_test.go` golden JSON updated for the added `redirect_chain` config field.

### Verification

```
$ go test -count=1 ./...
ok  	github.com/ffuf/ffuf/v2/pkg/ffuf	0.016s
ok  	github.com/ffuf/ffuf/v2/pkg/filter	0.010s
ok  	github.com/ffuf/ffuf/v2/pkg/input	0.011s
ok  	github.com/ffuf/ffuf/v2/pkg/output	0.012s
ok  	github.com/ffuf/ffuf/v2/pkg/runner	0.010s
$ go vet ./...
$ go build ./...
$ GOOS=windows GOARCH=amd64 go build ./...
```

Plus an end-to-end run against a local 3-hop server (`/admin → 301 /login → 302 /dashboard → 200`) confirms stdout, `-json`, `-of csv`, `-of md`, and `-of html` outputs all carry the chain.

### Naming

I went with `--redirect-chain` per the issue thread's framing. Open to `--show-redirects`, `--redirect-trace`, or always-on whenever `-r` is set — happy to rename or restructure if you'd prefer.


joohoi:
Hi, this sounds good! I'll review the code in a bit, but before that I have a design question:

I'm wondering the how the behavior plays (and what you would see as the idiomatic outcome from UX perspective) with options like raw-request logging `-od`; do/should all the raw request & response data be stored either as individual files, or clumped into one?

PR #905, opened 2026-07-15
Fix terminal control characters leaking into redirected output

### LINKED ISSUE #285: Ouput Issue
I m using ffuf with a for loop like this -> for url in ` cat $filename `; do ffuf -c -w path.txt -u $url/FUZZ -o result.json ; done >> result.txt
when I use cat result.txt it works fine        v1.2.0-git
________________________________________________

 :: Method           : GET
 :: URL              : https://www.google.com/FUZZ
 :: Wordlist         : FUZZ: path.txt
 :: Output file      : result.json
 :: File format      : json
 :: Follow redirects : false
 :: Calibration      : false
 :: Timeout          : 10
 :: Threads          : 40
 :: Matcher          : Response status: 200,204,301,302,307,401,403
________________________________________________

2018                    [Status: 301, Size: 239, Words: 9, Lines: 7]
2019                    [Status: 301, Size: 239, Words: 9, Lines: 7]
2017                    [Status: 301, Size: 239, Words: 9, Lines: 7]
2014                    [Status: 301, Size: 239, Words: 9, Lines: 7]
2015                    [Status: 301, Size: 239, Words: 9, Lines: 7]
2016                    [Status: 301, Size: 239, Words: 9, Lines: 7]

But when I use any text editor to open the file the response is 

[2K2018                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]
[2K2019                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]
[2K2017                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]
[2K2014                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]
[2K2015                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]
[2K2016                    [Status: [34m301[0m, Size: 239, Words: 9, Lines: 7]

any idea what to do

munzzyy:
# Description

`Progress`, `Info`, `Error`, `Warning`, `Raw` and every result line in
`pkg/output/stdout.go` unconditionally prepend the terminal line-clear
sequence (`\r\x1b[2K` on Linux/macOS, `\r\r` on Windows), and result
lines always append a trailing `\x1b[0m` color reset even when `-c`
was never passed. That's needed to redraw the progress bar on an
interactive terminal, but ffuf has no way to know it's not on one -
redirecting stdout/stderr to a file or a pipe just writes those raw
control bytes into the output instead.

Reproduced on current master:

```
$ ffuf -w wordlist.txt -u http://127.0.0.1:8899/FUZZ -mc 200 > out.txt
$ python3 -c "print(open('out.txt','rb').read())"
b'\r\x1b[2Ka                       [Status: 200, ...]\x1b[0m\n...'
```

Fix: track whether stdout/stderr are actual terminals (`os.ModeCharDevice`,
no new dependency) once in `NewStdoutput`, and only emit the clear-line
sequence and the color reset when writing to one. Explicit `-c` output
is unaffected in either case, since that's the user opting in
regardless of where the output goes.

After the fix, same command:

```
b'a                       [Status: 200, ...]\nb ...\nc ...\n'
```

Verified interactive behavior (progress bar redraw) is unchanged by
running the binary against a pty - the clear-line sequence is still
emitted there, only non-terminal output changes.

Fixes: #285

## Additonally

- [x] Added `munzzyy` to `CONTRIBUTORS.md`
- [x] Added a line to `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
     - Fix a bug in autocalibration strategy merging, when two files have the same strategy key
     - Fix a bug in -or, causing output to not to be written in any case
     - Fix panic when setting rate to 0 in the interactive console
+    - Fix terminal control characters being written to stdout/stderr when they're redirected to a file or pipe
   
 - v2.1.0
   - New
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -36,6 +36,7 @@
 * [l4yton](https://github.com/l4yton)
 * [lc](https://github.com/lc)
 * [mprencipe](https://github.com/mprencipe)
+* [munzzyy](https://github.com/munzzyy)
 * [nnwakelam](https://twitter.com/nnwakelam)
 * [noraj](https://pwn.by/noraj)
 * [oh6hay](https://github.com/oh6hay)
--- a/pkg/output/stdout.go
+++ b/pkg/output/stdout.go
@@ -51,7 +51,10 @@ func NewStdoutput(conf *ffuf.Config) *Stdoutput {
 }
 
 // isTerminal reports whether f is an interactive terminal, as opposed to a
-// file or a pipe.
+// regular file or a pipe. This also treats other character devices (e.g.
+// /dev/null) as terminals, which is harmless here since it only affects
+// whether a handful of extra control bytes get written to a destination
+// that isn't going to be read back anyway.
 func isTerminal(f *os.File) bool {
 	fi, err := f.Stat()
 	if err != nil {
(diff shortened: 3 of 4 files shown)


joohoi (changes requested):
Awesome, thanks! This looks good, just fix the linter nag, and we can merge.

munzzyy:
Fixed - the linter caught an unchecked io.Copy return in the test's capture helper, now it's handled with t.Fatalf like the pipe error right above it. Pushed.

joohoi (approved):
LGTM! Merging

PR #915, opened 2026-07-17
Fix interactive console being scrolled off screen by inflight results

joohoi:
Fix interactive console being scrolled off screen by inflight results

## Problem

When filters are open and matches stream fast, pressing ENTER to open the interactive console prints the banner but inflight requests keep completing and printing their result lines, scrolling the banner off screen.

`Job.Pause()` stops dispatch of new requests, but the up-to-`-t` workers already in flight keep going. Result printing happens in the worker goroutine (`Output.Result`), gated only by a `pauseCheckpoint()` that sits just before the print in `runTask`. A worker already past that checkpoint, or one doing slow replay-proxy / autocalibration I/O in the window before the print, still writes its line after the banner.

The existing `time.Sleep(500 * time.Millisecond)` before printing the banner was an attempt to let inflight work drain to the checkpoint. It adds latency yet cannot bound requests that take seconds, so it made the console feel unresponsive without fixing the scroll.

## Fix

Make the output provider pause-aware, so the gate lives at the layer that owns the terminal instead of depending on where each worker happens to be.

- `Stdoutput` gains `paused` and `pendingCount`, guarded by the existing `resultMutex`. `Result()` always records the match to `CurrentResults`; while paused it increments `pendingCount` and skips the live print, otherwise it prints as before.
- New `SetPaused(bool)` / `PendingResults() int` on the `OutputProvider` interface. Entering a pause resets the counter.
- Interactive ENTER now calls `SetPaused(true)` before `Job.Pause()`, then prints the banner immediately. The 500ms sleep is gone.
- Resume (ENTER, `resume`, `restart`) restores printing and reports `N new match(es) found while paused` so nothing looks lost. The prompt shows `[N new] >` while matches queue.

Held matches stay in `CurrentResults`, so `show`, `savejson`, `-o` output and the final results are unchanged. They are deliberately not replayed on resume, since that would re-stream exactly the noise the user paused to filter out.

Only `Result()` is suppressed. Errors and info stay visible, and the progress bar redraws one line in place (it does not scroll, and is quiescent while paused since the workers and dispatch loop are all blocked at the pause checkpoint).

## Limitations

This does not drain inflight requests, by design. A worker already past the pause checkpoint when ENTER is pressed can still print its line just after the banner. That window drops from request latency (up to seconds, a screenful) to a few instructions (at most a line or two), so the banner is no longer buried. Fully eliminating it would mean waiting for inflight requests to finish, which is exactly the unresponsiveness this avoids.

## Tests

- `pkg/output/stdout_pause_test.go`: while paused, `Result()` writes nothing to stdout, still records to `CurrentResults`, and increments `PendingResults()`; unpaused it prints and the counter stays 0; re-pausing resets the counter; 200 concurrent `Result()` calls while paused lose nothing and print nothing (run under `-race`).
- Full `go test -race ./...` green.

### DIFF
--- a/pkg/engine/autocalibration_test.go
+++ b/pkg/engine/autocalibration_test.go
@@ -36,8 +36,10 @@ func (o *NullOutput) FilterCurrentResults(keep func(ffuf.Result) bool) {
 	}
 	o.Results = filtered
 }
-func (o *NullOutput) Reset() {}
-func (o *NullOutput) Cycle() {}
+func (o *NullOutput) SetPaused(bool)      {}
+func (o *NullOutput) PendingResults() int { return 0 }
+func (o *NullOutput) Reset()              {}
+func (o *NullOutput) Cycle()              {}
 
 func TestAutoCalibrationStrings(t *testing.T) {
 	// Create a temporary directory for the test
--- a/pkg/ffuf/interfaces.go
+++ b/pkg/ffuf/interfaces.go
@@ -120,4 +120,10 @@ type Result struct {
 	ResultFile       string              `json:"resultfile"`
 	Host             string              `json:"host"`
 	HTMLColor        string              `json:"-"`
+	// Printed reports whether this result has already been shown to the user
+	// (streamed live, or surfaced in the "N new matches" summary on resume). It
+	// is the single source of truth for the interactive console's pending count,
+	// so filtering a held result out of CurrentResults also drops it from the
+	// count. Not serialized to any output format.
+	Printed bool `json:"-"`
 }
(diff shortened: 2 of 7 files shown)


joohoi:
Follow-up: two commits addressing correctness of held results when filters change mid-pause.

- Reconcile pending count with filter changes. The "N new matches found while paused" count was a separate counter that filter changes did not touch, so after adding a filter mid-pause the count could claim results that `show` no longer listed. The result's Printed flag is now the single source of truth; PendingResults counts the unprinted results in CurrentResults, so pruning a held result drops it from the count in the same operation. Leaving a pause marks held results seen, so a later pause only counts genuinely new matches.
- Fix in-console `fl` filter comparing against the wrong field: refreshResults built its probe with ContentLines set from ContentLength, so line-count filtering re-evaluated against byte length. Extracted resultProbe with the correct field, covered by a test.

Both covered by -race tests; full suite green.

PR #888, opened 2026-04-09
bump net to v0.52.0 and sys to v0.42.0

MiguelAAe:
# Description

fixes net CSV https://avd.aquasec.com/nvd/2023/cve-2023-39325/

## Additonally

- [x] If this is the first time you are contributing to ffuf, add your name to `CONTRIBUTORS.md`. 
The file should be alphabetically ordered.
- [x] Add a short description of the fix to `CHANGELOG.md`

Thanks for contributing to ffuf :)

PR #931, opened 2026-09-01
deps: bump golang.org/x/net to v0.58.0 — clears 9 reachable x/net/html advisories

Portll:
# Description

`golang.org/x/net` has been pinned at `v0.7.0` since early 2023. `govulncheck` reports **nine** advisories in `golang.org/x/net/html` that ffuf's own code reaches, all through one call site — `pkg/scraper/scraper.go:151` → `goquery.NewDocumentFromReader` → `html.Parse` — which runs on every response body under the default `-scrapers all`.

```
$ govulncheck ./...        # before
Your code is affected by 9 vulnerabilities from 1 module.

$ govulncheck ./...        # after
No vulnerabilities found.
```

The practical risk is a target that returns HTML engineered to hang or pin the parser (GO-2026-4441 infinite loop, GO-2026-4440 quadratic complexity, GO-2024-3333 non-linear parse). At the default `-t 40`, one hostile host can degrade a whole run.

**This is not urgent.** None of these is remotely exploitable against ffuf as a service — they need ffuf to parse hostile HTML, which is exactly what it is for. Worth doing because the parser fixes are real and because anyone running an SCA scanner over ffuf currently gets a KEV-flagged hit. See below on why that hit is misleading.

Fixes: none — no existing issue.

## Relation to #741 and #888

> **Corrected.** An earlier revision of this description said #741 "leaves 8 of the 9". That was wrong — it clears none of them. Corrected below, with the evidence, since it is a checkable number.

Every one of the nine advisories has a single affected range of the form `[0, fixed)`, per the Go vulnerability database:

| fixed in | advisories |
|---|---|
| 0.33.0 | GO-2024-3333 |
| 0.38.0 | GO-2025-3595 |
| 0.45.0 | GO-2026-4440, GO-2026-4441 |
| 0.55.0 | GO-2026-5025, GO-2026-5027, GO-2026-5028, GO-2026-5029, GO-2026-5030 |

So of the two open PRs that already bump x/net, neither clears the reachable set: **#741** (dependabot) targets 0.17.0, which is below every fix version and therefore clears **none of the nine**; **#888** (@MiguelAAe) targets 0.53.0, which clears four and leaves **five**, and currently conflicts with master. Clearing all nine requires ≥ 0.55.0. This branch is based on current master and goes to 0.58.0, the current release.

Happy to close this in favour of a refreshed #888 if that is preferred — the point is the target version, not who lands it.

Both of those PRs cite CVE-2023-39325 / CVE-2023-44487. For the record, **that one is not reachable from ffuf**: it is a `golang.org/x/net/http2` *server* defect, `go mod why golang.org/x/net/http2` answers *"main module does not need package"*, and ffuf runs no server. It is fixed here incidentally, not as the motivation.

## Cost: the Go floor moves 1.20 → 1.25

Clearing the GO-2026-502x/503x parser advisories needs x/net ≥ 0.55.0, and there is no version that fixes them and keeps 1.20.

Worth being precise about where that floor actually comes from, because it is not a cost of choosing 0.58.0: **`v0.51.0` is the first x/net release whose own `go.mod` requires `go 1.25.0`** (`v0.50.0` is still `1.24.0`). Every x/net from 0.51.0 up forces the same floor — including #888's 0.53.0 target. Any move to a current x/net pays this, whoever lands it.

Updated accordingly: `go.mod`, the `test.yml` matrix floor, the `golangci-lint.yml` setup-go pin, and the Go version stated in **both** `README.md` and `CONTRIBUTING.md`.

Three `t.Errorf("...: " + result)` calls in `pkg/ffuf/request_test.go` become `%s`. `go vet`'s non-constant-format-string check is an error once the go directive reaches 1.24, so the bump forces this rather than it being incidental cleanup.

## CI: golangci-lint v2, and node24 runtimes

The first revision flagged that it could not verify locally whether golangci-lint `v1.64.8` parses Go 1.25 export data. It does not — the lint job failed:

```
can't load config: the Go language version (go1.24) used to build golangci-lint
is lower than the targeted Go version (1.25.0)
```

v1 ended at v1.64.8, which is built with go1.24, so **no v1 release can lint this module**. The action moves to `golangci-lint-action@v9` with golangci-lint `v2.13.2`.

v2 changes two defaults, and the repo has no `.golangci.yml` to absorb them, so the bump would otherwise land **87 findings** as a lint failure. That set is identical on unpatched master, diffed line by line with zero divergence, so none of it is new here — measured at `33c67d2` originally and again at `23b5d68` after the rebase below, 87 uncapped and 32 capped on both trees each time. A new `.golangci.yml` holds both defaults at the v1 behaviour rather than fixing the findings, keeping this a toolchain change and not a lint-policy change:

- v1's default exclusions (`exclude-use-default: true`) became opt-in in v2 — restored via the four documented v1-equivalent presets.
- v2's `staticcheck` absorbed stylecheck (`ST*`) and quickfix (`QF*`), neither of which v1 ran by default — both disabled.

The resulting enabled set is `errcheck, govet, ineffassign, staticcheck, unused`: v1's default set, with gosimple folded into staticcheck as v2 requires. Fixing the 87 is a reasonable follow-up, but it is a different PR.

### Why that number is 87 and not 32

The same file also turns off golangci-lint's output truncation, which is worth calling out because it bit me while preparing this branch. Three separate defaults suppress findings, and the effect compounds silently:

| default | effect |
|---|---|
| `max-issues-per-linter: 50` | caps each linter |
| `max-same-issues: 3` | caps repeats of one message |
| `uniq-by-line: true` | drops extra issues sharing a line |

Out of the box, this tree reports **32 issues** when the real count is **87**. Nothing in the output indicates it truncated, so the printed number reads as the total — and the commit message on `1f8ae37` says 32 for exactly that reason, before I re-ran it uncapped.

The sharper problem is that a capped count is not stable enough to compare two trees with. Diffing master against this branch on default output gave **32 vs 32 with completely different composition**, because which duplicate instances survive the cap varies between runs. Only the uncapped run showed the two sets are line-for-line identical, which is the evidence behind "none of it is new here" above.

This changes nothing about the current result — the suite still reports 0 issues, since the exclusions are what remove the findings, not the caps. It only changes what a future failure is allowed to say, which for a CI gate is the number a contributor has to work from.

Separately, ahead of the runner deprecation, every action across the four workflows moves to its first major running node24:

| action | | |
|---|---|---|
| `actions/checkout` | v2, v3, v4 | → **v5** |
| `actions/setup-go` | v3, v5 | → **v6** |
| `golangci/golangci-lint-action` | v3 | → **v9** |
| `github/codeql-action` | v1 | → **v4** |
| `crazy-max/ghaction-import-gpg` | v6 | → **v7** |
| `goreleaser/goreleaser-action` | v6 | → **v7** |

Each was confirmed node24 by reading that ref's own `action.yml`, not from release notes — `golangci-lint-action@v8` is still node20, so v9 is the first that qualifies. Measured on the finished job logs: 5 `Node 20 is being deprecated` warnings before, 0 after, and the 4 `punycode` DeprecationWarnings went with the old action bundles.

One unrelated fix in the same job: `checkout` now runs **before** `setup-go`. It was harmless on setup-go v3, but v4+ caches by default and resolves `go.sum` from the working tree, which is empty if setup-go runs first.

This does mean the PR now touches `release.yml` and `codeql-analysis.yml`, which the dependency bump itself does not need. Happy to split those two files into their own PR if you would rather keep this one narrow.

## Checklist

- [x] Branch is based on the latest `master`. (3 ahead, 0 behind `23b5d68`. Rebased 2026-09-14 after #934–#936 landed; the only conflict was the `CHANGELOG.md` entry, which now sits under the unreleased `master` section rather than the since-tagged v2.2.0 block. `go test -race ./...`, `go vet`, `go mod tidy`, `go mod verify`, `govulncheck` and `golangci-lint v2.13.2` re-run clean on the rebased tree.)
- [x] `go test -race ./...` passes, and `gofmt` and `go vet` are clean. `go vet` clean; race suite green across all 12 packages. On `gofmt`: my changed files are clean. `gofmt -l .` does list 7 files under `pkg/input`, `pkg/interactive` and `pkg/output`, but that list is byte-identical on unpatched master, so it is pre-existing and I have left it alone rather than mixing a formatting sweep into a dependency bump.
- [x] Characterization goldens: no regeneration needed — `go test -run Characterization .` passes unchanged, since this alters no help, flag or config output.
- [x] First-time contributor / `CONTRIBUTORS.md` — deferring to the maintainers on whether this warrants an entry.
- [x] Add a short description of the change to `CHANGELOG.md`. (Dependency entry only; the CI change is not user-facing.)

Tested on darwin/arm64 with Go 1.27; not tested on the 1.25 floor itself or on Linux, though CI covers both on ubuntu and macOS. Also verified: `go mod verify` passes and `go mod tidy` is idempotent (no-op on both files).

On the module set — an earlier revision said it was "unchanged". More precisely: **nothing was added**, and five modules were *removed* from `go.sum` (`yuin/goldmark`, `x/crypto`, `x/mod`, `x/sync`, `x/xerrors`). All seven removed lines are `/go.mod` hashes only — module-graph metadata, no content hashes — and none were required in master's `go.mod`. They are pruned automatically once the go directive passes 1.17.


MiguelAAe:
Let's also update the golang linter version.

Portll:
> Let's also update the golang linter version.

updated for Node 24 and linted.

MiguelAAe (approved):
lgtm

PR #932, opened 2026-09-02
chore: repo hygiene — gofmt, workflow hardening, Dependabot, and the 87 parked lint findings

Portll:
# Description

Repo hygiene, found while preparing #931 and deliberately kept out of it. Twelve commits, each independently reviewable and most independently droppable. One non-cosmetic.

> **Stacked on #931.** This branch is based on `deps/xnet-0.58.0`, so the diff shown here includes that PR's three commits until it merges. The twelve commits below are the ones belonging to this PR — everything from `chore: gofmt` onward. Several of them touch files #931 also changes, so basing on `master` would have guaranteed conflicts rather than avoided them. Happy to rebase onto master if #931 is closed instead.

## Most important one:

**`fix: report write failures from the CSV, HTML and Markdown writers`**

All three had the same shape — create the file, `defer f.Close()`, write, return nil. A `Close` that fails on a file you have just written is how a short write surfaces: the data reached the kernel and not the disk. Discarding it means ffuf reports a successful `-o results.csv` over a file that is truncated or empty.

`writeCSV` had it twice, and the second is worse. `csv.Writer` buffers, and `defer w.Flush()` discarded the flush result — a `csv.Writer` reports write errors only via `w.Error()` after a flush. A disk filling up mid-results produced a silent partial CSV and a zero exit code.

These functions can now return an error where they previously returned nil. That is the point; the nil was wrong.

## Everything else

| commit | what |
|---|---|
| `chore: gofmt` | `gofmt -l .` listed 7 files on master; six need `//go:build` beside the legacy `// +build`. Verified constraints still resolve by building for windows/linux/darwin |
| `ci: add a Dependabot config` | Dependabot is active but unconfigured, so it only files security PRs (#741 is one). No version-update lane is why x/net sat at 0.7.0 from early 2023 |
| `ci: least-privilege GITHUB_TOKEN` | 3 of 4 workflows declared no `permissions:` at all and ran with the repo default |
| `ci: drop the 2020 CodeQL starter-template cruft` | `fetch-depth: 2` + `git checkout HEAD^2` is the old PR-head dance codeql-action has handled itself for years; the commented-out `make bootstrap && make release` names targets this repo does not have; the `git.io` links have been dead since 2022 |
| `ci: pin actions by commit SHA` | All 19 refs used a moving major tag. Every SHA verified to be the commit its named tag resolves to |
| `chore: make deliberately-ignored errors explicit` | The other 47 errcheck findings, where discarding is correct but nothing said so |
| `style: the six staticcheck findings that change no behaviour` | 2 redundant type declarations, 2 if/else ladders → tagged switches |
| `style: lowercase error strings (ST1005)` | **The only commit that changes what a user sees.** See below |
| `ci: stop excluding what is now clean` | Drops the suppression `.golangci.yml` carried, now that the findings are fixed rather than excused |
| `deps: bump adrg/xdg to v0.5.3` | Moves `gopkg.in/yaml.v3` off a version inside CVE-2022-28948 |
| `docs: changelog` | The three user-visible changes |

## The lint count, and why it is 87

#931 added a `.golangci.yml` that held the linter policy at v1's defaults so a toolchain bump would not smuggle in a policy change. That parked 87 findings rather than fixing them. This branch fixes them and removes the parking.

 golangci-lint's default output says **32**. 

Three separate defaults suppress findings and the effect compounds silently — `max-issues-per-linter: 50`, `max-same-issues: 3`, `uniq-by-line: true`. Nothing in the output indicates it truncated. The final `.golangci.yml` turns all three off, so a future failure reports its real size.

Config is verified as a ratchet: injecting a single unused function returning `fmt.Errorf("Capitalized with a stop.")` makes golangci-lint exit 1 with three findings, and removing it returns exit 0.

## On the error-string commit

`style: lowercase error strings` is split out on purpose and nothing after it depends on it, so it can be dropped on its own if you would rather ffuf's messages stayed capitalised.

29 lose a leading capital and 2 lose a trailing full stop. `"FUZZ keyword defined, but we are using sniper mode."` only loses the stop — FUZZ is a keyword, not a sentence opener. Mid-string capitals are untouched, so `"Expected http, https or socks5 url"` reads as before.

Two consumers moved with it: `pkg/ffuf/optionsparser_test.go` asserts the proxy error text against a hardcoded literal, and `testdata/config_error_missing_wordlist.golden` captures the `-w` error. I found those by checking **all 22** distinct strings this branch changes against the test and golden tree — an initial spot-check of six missed the proxy pair, and the race suite caught it rather than review.

## Deliberately not done

- **The other nine outdated modules.** `goquery v1.8.0 → v1.13.0` is five minor versions of parser behaviour and deserves its own review. With the Dependabot config in this branch, these now arrive one at a time on a schedule — doing them by hand here would pre-empt that badly.
- **Windows in the test matrix.** `test.yml` already documents why it is absent: the help-golden test deadlocks on Windows' small stdout pipe buffer and a couple of output tests assume POSIX terminal control. That needs the tests fixed, which is a change, not cleanliness.

## Checklist

- [x] Branch is based on the latest `master` (via #931; 0 behind).
- [x] `go test -race ./...` passes, and `gofmt` and `go vet` are clean. All three exit 0, checked on the exit code rather than on filtered output. The `-tags=timing` suite exits 0 too. `gofmt -l .` is now empty, where it listed 7 files on master.
- [x] Characterization goldens: `config_error_missing_wordlist.golden` regenerated with `go test -run Characterization -update-golden .`; re-running it leaves the tree clean.
- [x] First-time contributor / `CONTRIBUTORS.md` — deferring to the maintainers.
- [x] Added to `CHANGELOG.md` — the three user-visible entries only.

Also verified: `golangci-lint run` exits 0, `govulncheck ./...` reports no vulnerabilities, and the binary cross-builds for windows, linux and darwin. Tested on darwin/arm64 with Go 1.27.


Portll:
Rebased onto the rebased #931 (which now sits on `23b5d68`), so this is 0 behind master again. The only conflict was the CHANGELOG entry, which moved with #931's into the unreleased `master` section. Re-run on the rebased branch: `go mod tidy` no-op, `go vet` clean, `gofmt -l .` empty, golangci-lint 2.13.2 0 issues, `go test -race ./...` green across all packages.

